PR #1, opened 2024-01-29
RFC: Establish a formalized RFC process

beta-ziliani:
Preview: https://github.com/crystal-lang/rfcs/blob/feat/add-readme/README.md

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -1,0 +1,87 @@
+- Feature Name: (fill me in with a unique ident, `my_awesome_feature`)
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [crystal-lang/rfcs#0000](https://github.com/crystal-lang/rfcs/pull/0000)
+- Issue: [crsytal-lang/crystal#0000](https://github.com/crystal-lang/crystal/issues/0000)
+
+# Summary
+
+One paragraph explanation of the feature.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Guide-level explanation
+
+Explain the proposal as if it was already included in the language and you were teaching it to another Crystal programmer. That generally means:
+
+- Introducing new named concepts.
+- Explaining the feature largely in terms of examples.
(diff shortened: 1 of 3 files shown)


nobodywasishere (approved):
Will currently existing PRs be required to go through this process?

straight-shoota:
@nobodywasishere I think it mostly depends where the process is at. When the course of action is already pretty much established and implementation just needs some refinements, there's no need to launch a formalized RFC process.
The idea of this tool is to bring more clarity about where proposals stand and what they need to get finalized and approved.
So it could be useful for some currently pending PRs to write a formalized RFC here. I wouldn't start with any of that unless some explicitly asks you to.

We're just beginning with this new mechanism and need to figure out some of the details along the way. Of course, this should not add too much bureaucracy to hinder progress, but help it forward with a bit more formalized process.

RX14 (approved):

beta-ziliani:
What I would like to see (without any rush!) is for old and heavily debated PRs/RFCs to land here in a well-structured manner, incorporating the feedback from the discussions.

straight-shoota:
@devnote-dev I'm not sure how exactly the question is meant, but I'm going to assume you're asking about ongoing RFC discussions on the forum such https://forum.crystal-lang.org/t/rfc-within-a-block/417 ?
The forum is a good place to start talking about a topic in a more informal way with a broad perspective. It helps gauge community opinions and narrow done a path when there are plenty of options. I think this is still very useful for a perliminary exploration and once it has reached a certain point of concretness, can result in a formal RFC which woulnd't start at 0 and build upon the forum discussion.

devnote-dev:
@straight-shoota Yes that's what I meant, thank you for clarifying!

devnote-dev (approved):

HertzDevil:
Probably too late to the party, but I don't think the template should use multiple h1 titles and instead they should be h2

straight-shoota:
Those documents are not set in stone. We can just fix it. Trivial syntax changes won't need a formal RFC process, of course ;)

PR #2, opened 2024-02-05
RFC 0002 - MT Execution Contexts

ysbaddaden:


beta-ziliani:
Preview: https://github.com/crystal-lang/rfcs/blob/rfc-0002-mt-execution-contexts/0002-execution-contexts.md

0002-execution-contexts.md
```diff
@@ -0,0 +1,334 @@
...
+- **Thread**: program whose execution is controlled by the operating system ([wikipedia](https://en.wikipedia.org/wiki/Thread_(computing))).
+- **Fiber**: unit of works or “functions whose execution you can pause” ([wikipedia](https://en.wikipedia.org/wiki/Coroutine)), we can run multiple fibers per thread.
+- **Scheduler**: manages fibers’ executions inside the program (controlled by Crystal), unlike threads that are scheduled by the operating system (outside of the accessibility of the program).
+- **Event Loop**: an abstraction to wait on specific events, such as timers (e.g. wait until a certain amount of time has passed) or IO (e.g. wait for an socket to be readable or writable).
+- **Multi-threaded (MT)**: the environment uses multiple threads, fibers can run both concurrently and in parallel on different CPU cores.
+- **Single-threaded (ST)**: the environment uses only one thread, fibers can never run in parallel but still run concurrently.
+
+## Concurrency
+
+> I only quickly sum up the concurrency model. There are very interesting topics about concurrency to consider (e.g. structured concurrency, fiber cancellation), but this RFC focuses on parallelism.
```
straight-shoota, line 25: **suggestion:** GitHub Flavoured Markdown supports an admonition syntax s: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

```suggestion
> [!NOTE]
> I only quickly sum up the concurrency model. There are very interesting topics about concurrency to consider (e.g. structured concurrency, fiber cancellation), but this RFC focuses on parallelism.
```

straight-shoota:
The distinction between execution context and scheduler could need a bit refinement. There is definitely some overlap in functionality, just by comparing the API. I guess execution contexts might take over some features of the current scheduler?

0002-execution-contexts.md
```diff
@@ -0,0 +1,475 @@
...
+
+# Unresolved questions
+
+## Introduction of the feature
+
+The feature will likely require both the `preview_mt` and `execution_contexts` flags, at least for the initial implementations.
+
+Since it is a breaking change from stable Crystal, unrelated to `preview_mt`, MT and execution contexts may only become default with a Crystal major release?
+
+Maybe MT could be limited to 1 thread by default, and `spawn(same_thread)` be deprecated, but what to do with `spawn(same_thread: true)`?
```
RX14, line 463: Pass it to the EC and let it raise if not compatible. The default single threaded context could support this but any MT context that does work stealing would raise. Deprecate the parameter and schedule it for removal in 2.0.

0002-execution-contexts.md
```diff
@@ -0,0 +1,475 @@
...
+
+## Default context configuration
+
+This proposal doesn’t solve the inherent problem of: how can applications configure the default context at runtime (e.g. number of MT schedulers) since we create the context before the application’s main can start.
```
RX14, line 467: This proposal supports creating multiple execution contexts, let the application configure it's own EC and start fibers in it if required. That allows all the complexity the app needs when configuring the context the application actually runs in, because it's initialized by the application. The root context does not have to be well-used by the application.

RX14 (commented):
Absolutely lovely, well-written proposal. I completely agree with the design intent here, and only have a few—mostly overlapping—comments about event loops and the default context. The vast majority of this design is exactly what I would like to see in crystal.

0002-execution-contexts.md
```diff
@@ -0,0 +1,475 @@
...
+**Isolated Context**: only one fiber is allowed to run on a dedicated thread (e.g. `Gtk.main` or game loop), thus disabling concurrency on that thread; an issue is what happens if the fiber tries to spawn (raise? forward to default context?), triggers the event-loop (IO, timers), or tries to communicate with other contexts? An advantage is that the fiber can own a thread, without blocking other fibers.
+
+Precisions:
+
+- The above list isn’t exclusive: there can be other contexts with different rules (for example MT without work stealing).
+- Each context isn’t exclusive: an application can start as many contexts in parallel as it needs.
+
+## The default execution context
+
+Crystal would start a MT execution context with work-stealing where fibers are spawned by default. The goal of this context is to provide an environment that should fit most use cases to freely take advantage of multiple CPU cores, without developers having to think much about it, outside of protecting concurrent accesses to a resource or, preferably, using channels to communicate.
```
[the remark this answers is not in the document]
ysbaddaden: There are more and more cores (even on x86) so limiting to one by default feels more and more dated.

I think it's beneficial to the Crystal ecosystem (shards) to know that they live in a parallel world _by default_ and it's easier & quicker to notice parallel issues, instead of a dual "usually ST" vs "sometimes MT".

0002-execution-contexts.md
```diff
@@ -0,0 +1,475 @@
...
+# Reference-level explanation
+
+An execution context shall provide:
+
+- configuration (e.g. number of threads, …);
+- methods to spawn, enqueue, yield and reschedule fibers within its premises;
+- a scheduler to run the fibers (or many schedulers for a MT context);
+- an event loop (IO & timers):
+
+  => this might be complex: I don’t think we can share a libevent across event bases? we already need to have a “thread local” libevent object for IO objects as well as for PCRE2 (though this is an optimization).
```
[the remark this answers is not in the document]
ysbaddaden: We currently have one per scheduler. I was thinking about one per EC, but you may be right: one per process might be enough and better :thinking:

0002-execution-contexts.md
```diff
@@ -0,0 +1,475 @@
...
+
+The **Tokio** crate for **Rust** provides a single-threaded scheduler or a multi-threaded scheduler (with work stealing), but it must be chosen at compilation time. Apparently the MT scheduler is selected by default (to be verified).
+
+# Unresolved questions
+
+## Introduction of the feature
+
+The feature will likely require both the `preview_mt` and `execution_contexts` flags, at least for the initial implementations.
+
+Since it is a breaking change from stable Crystal, unrelated to `preview_mt`, MT and execution contexts may only become default with a Crystal major release?
```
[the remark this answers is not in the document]
ysbaddaden: Default EC doesn't have to be ST: it can be MT:1 and the app could grow it :thinking:

ysbaddaden:
@RX14 I applied your suggestions to the RFC.

0002-execution-contexts.md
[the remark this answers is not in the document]
straight-shoota: Yeah of course we cannot simply rollout a change to run Crystal with multiple threads by default. That would be considered a breaking change. Many libraries are not thread-safe and would break on this.

So I think this document describes the desired state we want to reach eventually. We'll have to put fruther thought into how to get there and design the migration path.

RX14:
@ysbaddaden I think `-Dmt` is probably not necessary, and the exact implementation plan for crystal 2 is best left deferred until there's operational experience, but I agree on everything else.

I envision the root execution context being MT or ST a moot point, because every well-architected app has a single `App.run` line at the top-level and converting that top-level code to be spawning a MT context and waiting for that fiber should be a one-liner if we have the right helper methods in place.

If we all agree, maybe we can start on the other 90% of the RFC: bikeshedding naming. I like `ExecutionContext::Parallel`, because I don't like the idea of implementation details (threads) leaking into the name.

ysbaddaden:
@RX14 The `mt` flag may not be necessary in Crystal v1, as the default context could be MT:1 and resized on demand (still no breaking change). I'll still push for MT:N to be the default in Crystal v2. Execution contexts are a mean to _further_ control the parallelism in very specific cases, not the end solution. I believe developers shouldn't have to care about it until you have to.

I wouldn't bikeshed the namings just yet. As I'm experimenting with the types, I feel that the difference is getting thinner and thinner. In fact, Kotlin only has a single scheduler implementation, and a couple constructors to start execution contexts with 1 (ST) or many threads (MT).

I'm also struggling with the inheritance: `EC::MT < EC` makes sense, but so does `EC::MT::Scheduler < EC` as we want `EC.current` to point to the current MT scheduler running on the thread, not the shared MT context (it's easier to reach the context from the scheduler).

ysbaddaden:
I forgot again but MT:1 would break `spawn(same_thread: true)` in Crystal 1. It's a NOOP without the `preview_mt` but the parameter was still exposed in the public API :sob:

straight-shoota:
I think we can accept breakage with `same_thread: true`. It only works with `preview_mt` which is explicitly a preview feature, so there should be no expectations on compatibility in a setting outside of `preview_mt`.

0002-execution-contexts.md
```diff
@@ -0,0 +1,493 @@
...
+5. Different contexts could have different priorities, maybe allowing the operating system to allocate threads more effectively in big.LITTLE architectures.
+
+# Reference-level explanation
+
+An execution context shall provide:
+
+- configuration (e.g. number of threads, …);
+- methods to spawn, enqueue, yield and reschedule fibers within its premises;
+- a scheduler to run the fibers (or many schedulers for a MT context);
+- an event loop (IO & timers):
```
yxhuvud, line 225: It probably needs to be mentioned that it needs to continue to work with channels and mutexes. This is *somewhat* straightforward today with how fibers are bound to a thread once executed, so the simple version is to just schedule it using the normal interfaces. But interfaces for scheduling are not necessarily the same between execution contexts as they are within it! 

For example, a basic work stealing scheduler can default to just enqueue a fiber in the executing scheduler/thread context, and let distribution between threads happen in other ways. But that doesn't work if the thing to wake up doesn't live in the same context - then it needs to be communicated somehow. And then the question is what to communicate it to.

Also somewhere it should probably be defined what happens if channel interaction happens in an isolated context (as defined above).

0002-execution-contexts.md
```diff
@@ -0,0 +1,493 @@
...
+    current = self.current
+    if fiber.execution_context == current
+      current.resume(fiber)
+    else
+      raise "BUG"
+    end
+  end
+
+  # the following methods can be called from whatever context and must thrad
+  # safe (even ST):
```
yxhuvud, line 323:
```suggestion
  # the following methods can be called from whatever context and must be thread
  # safe (even ST):
```

0002-execution-contexts.md
```diff
@@ -0,0 +1,493 @@
...
+  class MultiThreaded < ExecutionContext
+    getter name : String
+
+    def self.new(name : String, exactly : Int32)
+      new(name, exactly, exactly)
+    end
+
+    def initialize(@name : String, @minimum : Int32, @maximum : Int32)
+      # todo: start @minimum threads
+    end
```
yxhuvud, line 357: Allowing a dynamic amount of threads requires more synchronization and complexity than having a static amount. While it  sounds nice to be able to adjust, it probably warrants its own separate class. Making certain all threads are in a waiting state before starting to actually queue stuff allows a bunch of simplifications with less mutexes and risks a lot fewer possible race conditions too.

0002-execution-contexts.md
```diff
@@ -0,0 +1,493 @@
...
+3. Drop the `preview_mt` flag and make execution contexts the only compilation mode (at worst introduce `without_mt`):
+   - one can limit the default context to a single thread to (disabling parallelism at runtime);
+   - sync primitives must be thread-safe, because fibers running on different threads or in different execution contexts will need to communicate safely;
+   - sync primitives must be optimized for best possible performance when there is no parallelism;
+   - community maintained shards may propose alternative sync primitives when we don’t want thread-safety to squeeze some extra performance inside a single-threaded context.
+
+> [!NOTE]
+> The breaking changes can be postponed to Crystal 2 by making the default execution context to be ST, keep supporting `same_thread: true` for ST while MT would raise, and `same_thread: false` would be a NOOP.
+>
+> A compilation flag can be introduced to change the default context to be MT in Crystal 1 (e.g. keep `-Dpreview_mt` but consider just `-Dmt`).
```
yxhuvud, line 431: Yet another option is to simply flip the default for a couple of versions (like 2.x), while allowing opting out of mt and new semantics given that flag

Blacksmoke16:
As someone who isn't familiar with this stuff at all, my random question is:

> Do we need to do anything in relation to like how Intel has P and E cores now? Like as a way to signal to the OS's thread scheduler that a fiber should have a preference on where it runs? Or is that something the OS itself handles somehow?

0002-execution-contexts.md
```diff
@@ -0,0 +1,493 @@
...
+
+def sleep(time : Time::Span) : Nil
+  ExecutionContext.sleep(time)
+end
+```
+
+And the proposed API:
+
+```crystal
+abstract class ExecutionContext
```
yxhuvud, line 289: What is the lifetime of an execution context? Do we need some way to signal to it that it can close down when the last fiber has finished?

0002-execution-contexts.md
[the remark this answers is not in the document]
ysbaddaden: Yeah, I'm wondering about that. We currently have one libevent base per thread, and that overcomplexifies IO::Evented with lots of objects to allocate in the HEAP for each IO and thread.

I'm probably being naive (though Go seems to do that) but maybe a global event loop wouldn't behave so badly? Even with the potential contention on adding an event to the libevent base, especially on machines with too many cores to count (e.g. [ARM Neoverse](https://en.wikipedia.org/wiki/ARM_Neoverse)).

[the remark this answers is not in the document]
yxhuvud: On the other hand not having to synchronize everything and their mother make other things a lot less complex. It is quite liberating to not give a shit about what other threads are doing. 

FWIW, having one dedicated ring per thread is also how the makers of io_uring recommend multi thread usage.

0002-execution-contexts.md
ysbaddaden: No difference from today: the default EC hijacks the main thread and doesn't run the scheduler's run loop on that thread, so when the main thread's main fiber terminates that terminates the main thread and thus the process.

EC still need explicit sync (e.g. channel, wait group, nurseries, ...).

The shutdown procedure I mentioned is about stopping an EC, that would ask the threads to stop (cooperative) but we might ask to stop while there are runnable fibers (or fibers that will be reenqueued) and those should be cancelled, and... that sounds like something for structured concurrency (hence future evolution).

ysbaddaden:
@Blacksmoke16 From what I read specifying a thread priority can hint the OS to schedule the thread on a big (efficient) or little (power) core. We can also set a thread affinity to a given core, but we must detect the core type beforehand.

0002-execution-contexts.md
yxhuvud: > the default EC hijacks the main thread and doesn't run the scheduler's run loop on that thread

Yes, this works for the main thread, and depends on the OS to do cleanup. I was more thinking of the nested case when someone creates a new execution context. How would the API for actual usage look like? Perhaps the lifecycle for children fall out of that?

ysbaddaden: > Perhaps the lifecycle for children fall out of that?

Yes, I think it should be it's own type. Maybe a wrapping Execution Context? which brings an interesting point to the API: allow wrapping.

For example my current experiment assigns an EC::Scheduler < EC as the current EC, so wrapping would need to wrap both the EC (public interface) and EC::Scheduler (thread local interface) to monitor fibers :thinking:

0002-execution-contexts.md
ysbaddaden: Enqueue doesn't have to need much sync. Go pushes to a bounded local queue (per scheduler) with overflow to a global queue; threads can be started at any time: when they reach the run loop they will grab a batch of fibers from the global queue or steal from another scheduler. Stopping ain't more complex, schedulers aren't tied to a specific thread, the thread detaches the scheduler and returns itself to the thread pool.

The complexity is more in _when_ to start / stop a thread.

0002-execution-contexts.md
ysbaddaden: I added a couple notes about the ability the wrap an EC to add further capabilities (e.g. monitoring child fibers).

0002-execution-contexts.md
ysbaddaden: And that complexity pushes for it to become a "future evolution".

0002-execution-contexts.md
```diff
@@ -0,0 +1,508 @@
...
+
+1. We can have fully isolated single-threaded execution contexts, mimicking the current MT implementation.
+
+2. We can create an execution context dedicated to handle the UI or game loop of an application, and keep the threads of the default context to handle calculations or requests, never impacting the responsiveness of the UI.
+
+3. We can create an MT execution context for CPU heavy algorithms, that would block the current thread (e.g. hashing passwords using BCrypt with a high cost), and let the operating system preempt the threads, so the default context running a webapp backend won't be blocked when thousands of users try to login at the same time.
+
+4. The crystal compiler doesn’t need MT during the parsing and semantic phases (for now); we could configure the default execution context to 1 thread only, then start another execution context for codegen with as many threads as CPUs, and spawn that many fibers into this context.
+
+5. Different contexts could have different priorities and affinities, to allow the operating system to allocate threads more effectively in big.LITTLE architectures.
```
straight-shoota, line 217: **nitpick:** "big.LITTLE" is just one particular instance of a heteogenous computing architecture, I think it's better to use the generic term here to include all kinds of forms.
Also it should probably be "efficient", not "effective".

```suggestion
5. Different contexts could have different priorities and affinities, to allow the operating system to allocate threads more efficiently in heterogenous computing architectures.
```

0002-execution-contexts.md
```diff
@@ -0,0 +1,508 @@
...
+  # the default execution context (always started)
+  class_getter default = MultiThreaded.new("DEFAULT", minimum: 1, maximum: System.cpu_count.to_i)
+
+  def self.current : ExecutionContext
+    Thread.current.execution_context
+  end
+
+  # the following methods delegate to the current execution context, they expose
+  # the otherwise protected instance methods which are only safe to call on the
+  # current execution context:
```
straight-shoota, line 303: **question:** This part of the API feels a bit itchy. The methods are exposed publicly, but have a strict prerequisite: "only safe to call on the current execution context". I believe this is void for `.reschedule` because it explicitly targets `current`. So the only weird kid is `.resume`.

I'm not sure if you've given much thought into these peculiarities or just tried to map features of the current API to the new model as directly as possible.
Perhaps we might want to reconsider the implementation of `Fiber#resume`.
The general meaning of this method is to stop executing the current fiber and give control to the targeted fiber.
If they share the same EC, the targeted fiber can (potentially) replace the current one directly. This should be the choice of the EC, though.
If they are in different ECs, the semantics could be reached by enqueing the target fiber and stopping the current one.
```cr
  # Resumes `fiber` in the execution context. Raises if the fiber
  # doesn't belong to the context.
  def self.resume(fiber : Fiber) : Nil
    if fiber.execution_context == current
      current.resume(fiber)
    else
      fiber.execution_context.enque(fiber)
      current.reschedule
    end
  end
  ```

File: 0002-execution-contexts.md (hunk 5 above)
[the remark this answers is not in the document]
Comment (yxhuvud) [reply 0002-execution-contexts.md:228]: Yes, storing it on the stack is a very useful technique for io_uring as well (to keep it alive during submission). I've made a lot of use of that and it helped a lot. There the completion event need to be handled somehow though - I stored it in the fiber itself until it was awakened and could process it, but there may be better techniques.

One good thing about the current implementation though is that it handles the thundering herd problem decently, as far as I've been able to see. That is, if multiple fibers are waiting for something only one of them will wake when something happens. With many listeners that may become something to keep track of.

File: 0002-execution-contexts.md (hunk 13 above)
Comment (ysbaddaden) [reply 0002-execution-contexts.md:303]: I didn't give it much thought because:

1. it's not used anywhere in the stdlib, and only one spec relies on it; I'd go for deprecating `Fiber#resume` and drop it in Crystal 2;
2. it's already not working correctly with the current MT. It even used to be happily resumed by any thread :shrug:

I only got it "fixed" recently by raising on thread mismatch. It's probably in 1.11

Comment (straight-shoota) [reply 0002-execution-contexts.md:303]: Sounds good. Let's add `@[Deprecated]` then to document this.

File: 0002-execution-contexts.md [hunk 14]
```diff
@@ -0,0 +1,475 @@
...
+
+**Single Threaded Context**: fibers will never run in parallel, they can use simpler and faster synchronization primitives internally (no atomics, no thread safety) and still communicate with other contexts with the default thread-safe primitives; the drawback is that a blocking fiber will block the thread and all the other fibers.
+
+**Multi Threaded Context**: fibers will run in parallel and may be resumed by any thread, the number of schedulers and threads can grow or shrink, schedulers may move to another thread (M:N schedulers:threads) and steal fibers from each others; the advantage is that fibers that can run should be able to run, as long as a thread is available (i.e. no more starving threads) and we can be shrink the number of schedulers;
+
+**Isolated Context**: only one fiber is allowed to run on a dedicated thread (e.g. `Gtk.main` or game loop), thus disabling concurrency on that thread; an issue is what happens if the fiber tries to spawn (raise? forward to default context?), triggers the event-loop (IO, timers), or tries to communicate with other contexts? An advantage is that the fiber can own a thread, without blocking other fibers.
```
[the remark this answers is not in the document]
ysbaddaden: You're right. I previously updated the RFC to reflect that:

1. The event loop can run normally: it suspends the current fiber waiting for the event and that's it.
2. We can specify another context to spawn fibers into (can be the default context by default).

0002-execution-contexts.md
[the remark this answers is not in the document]
ysbaddaden: Exactly; a RFC should reflect the _desired_ state.

0002-execution-contexts.md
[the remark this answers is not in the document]
ysbaddaden: We shall move the "refactor the event loop" into a proper issue on https://github.com/crystal-lang/crystal/issues

It doesn't need a RFC as it's mostly internal implementation detail.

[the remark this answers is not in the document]
ysbaddaden: I might be interesting to consider one event loop (EL) per execution context (EC) :thinking:

With a single EL per EC, a starving scheduler will run the EL and enqueue every resumable fibers for the EC.

With one EL per scheduler, a starving scheduler will run its own EL and push the fraction of fibers it happened to have, possibly flooding the EC queues and delaying other schedulers from running their own EL, and delaying resumable fibers from being resumed.

That may be a reason why Go has a single EL: it could be unfair otherwise.

ysbaddaden:
@RX14 I growing fond of `Parallel`.
I think `Exclusive` could be better and more explicit than `Isolated`?
I'm not sure about the single threaded context... or maybe `Single`?

ysbaddaden:
After spending 2 days chasing after a bug, I realized I mixed up two different things into a single ExecutionContext class: the **execution context** (cross context) and the **fiber scheduler** (thread local) which led to a nasty bug (cross scheduler local enqueues).

I'm trying to separate these into 2 different types (e.g. `ExecutionContext` and `ExecutionContext::Scheduler`), and probably make them as modules, so we can have a single type implementing both modules (e.g. `Isolated`) or a couple types each implementing one module (e.g. `MultiThreaded`).

text/0002-execution-contexts.md
```diff
@@ -0,0 +1,571 @@
...
+  # todo: implement abstract methods
+end
+```
+
+A multithreaded context should implement both modules as different types, since
+there will be only one execution context but many threads that will each need a
+scheduler. For example:
+
+```crystal
+  class MultiThreaded < ExecutionContext
```
Sija, line 399: Is this still relevant?

```suggestion
  class MultiThreaded
```

text/0002-execution-contexts.md
```diff
@@ -0,0 +1,571 @@
...
+    end
+
+    def spawn(name : String?, same_thread : Bool, &block) : Fiber
+      raise RuntimeError.new if same_thread
+      @spawn_context.spawn(name, &block)
+    end
+
+    # todo: prevent enqueue/resume of anything but @fiber
+  end
+end
```
Sija, line 447:
```suggestion
class Isolated < SingleThreaded
  def initialize(name : String, @spawn_context = ExecutionContext.default, &@func : ->)
    super name
    @fiber = Fiber.new(name: name, &@func)
    enqueue @fiber
  end

  def spawn(name : String?, &block) : Fiber
    @spawn_context.spawn(name, &block)
  end

  def spawn(name : String?, same_thread : Bool, &block) : Fiber
    raise RuntimeError.new if same_thread
    @spawn_context.spawn(name, &block)
  end

  # todo: prevent enqueue/resume of anything but @fiber
end
```

ysbaddaden:
Thinking again about names:

- ST: Mono? Single? Concurrent (to oppose on Parallel)?
- MT: Parallel is still the best for MT :+1: 
- I prefer Isolated over Exclusive after all (the fiber gets isolated);

I'm also thinking about simple constructors:

```crystal
io_workers = ExecutionContext.concurrent
cpu_workers = ExecutionContext.parallel(size: 8)
ui = ExecutionContext.isolate { UI.main_loop }
```

Yet, I'm still struggling for a nice name to the single threaded context :disappointed:

straight-shoota:
`SingleThreaded` maybe?

I'm not sure we need such convenience constructors. This isn't essential anyway and we can figure it out later.

ysbaddaden:
@straight-shoota yes, the convenience fonctions aren't needed, but while `ExecutionContext::Parallel` and `ExecutionContext::Isolated` feel nice, `ExecutionContext::SingleThreaded` doesn't have the same catchy feeling.

Qard:
Just wanted to share some thoughts/ideas I'm using in my own Fiber-based multi-threaded language.

There are two distinct access patterns that can be isolated into different types of Fiber construction:

- In the typical case it is desirable that a Fiber takes access to variables present in the lexical scope in which it is constructed. As such, distributing to another thread creates synchronization challenges which make doing so undesirable.
- In _some_ cases you have a singular "task" you want distributed which it would be reasonable that the processor takes ownership of that in the sense that a language like Rust does.

How I chose to handle this is with two distinct Fiber types: a local Fiber which can access any in-scope data but can only execute on the thread in which it was created, and an ownership-claiming Fiber which specifically _avoids_ running on the thread in which it was created, taking ownership of a passed in value and bringing that over with it to whatever thread it gets dispatched to.

On top of that I have a work-stealing mechanism which can pull a Fiber _back_ to its origin thread when it's not busy, but the Fiber would otherwise be guaranteed to run on a different thread.

This design focuses very strongly on maintaining locality to achieve the best performance and to avoid synchronization complexity as much as possible while making distribution over threads certain when loop utilization would benefit from it.

In a typical app server you're going to see http requests come in which are largely isolated from each other and can be immediately distributed to other threads, but then the processing within those tends to need to share more with each other, and what you need within the scope of _parts_ of that request is more likely just concurrency and not actually parallelism. If you _need_ parallelism you can reach for that tool, but if a more restrictive tool is provided to ensure actual isolation of the work for that Fiber then it can be much more effectively distributed.

Another thing you see often with http requests is that _most_ object references within a request are to objects allocated in that request, but there are a small number of exceptions which typically fit neatly into a shared interface category such as a database connection (or pool) or an app server exposing utilities to each route. For these types I opted to build actors into my language and have any interaction with these objects go through a shim copied into the ownership-taking fibers which dispatch these interactions across threads through Fiber yields as the "lock" mechanism. A task simply gets passed over to the thread owning the object to do that interaction and the result of that interaction resolves the yield on the other end.

In my language I went for an explicit form of this actor type specialization, but you _could_ also go for explicit construction of a proxy type when passing to an ownership-taking Fiber.

In any case, what I wanted to convey is that _in my experience_ there are two distinct ways in which people want to distribute work with Fibers and isolating those two behaviours can actually have a lot of benefits. 🙂

ysbaddaden:
The current constructors are rather complex and inconvenient to type, and pushing `ExecutionContext` under the `Fiber` namespace didn't help. I mean:

```crystal
ctx = Fiber::ExecutionContext::Isolated.new(name) { }
ctx = Fiber::ExecutionContext::SingleThreaded.new(name)
ctx = Fiber::ExecutionContext::MultiThreaded.new(name, size: 1..4)
```

I'm wondering if we could have simpler helpers. For example (thinking out loud, don't take things for granted):

```crystal
ctx = Fiber.start_isolated(name) { } # or #spawn_isolated ?
ctx = Fiber.start_concurrent_context(name)
ctx = Fiber.start_parallel_context(name, size: 1..4)
```

I'm not sure it's really better. Maybe a bit more explicit thanks to the `start_` prefix?

The naming might also indicate that maybe the ST and MT classes should be named `Concurrent` and `Parallel` of `SingleThreaded` and `MultiThreaded`?

ysbaddaden:
I updated the phrasing and names of the proposed execution contexts.

Reading the RFC again, I believe that its purpose, to propose a new API to achieve parallelism in Crystal, can be considered done, even though the feature isn't fully released.

I believe we can archive RFC 0002, and start a new RFC with details on the actual execution contexts (Concurrent, Parallel, Isolated), what guarantees they bring and how they differ, and what still needs to be done: resize, shutdown, thread pool and detach on syscall, mostly.

text/0002-execution-contexts.md
```diff
@@ -393,43 +409,47 @@ class SingleThreaded
...
 ```
 
-Finally, an isolated context could extend the singlethreaded context, taking
-advantage of its —in practice we might want to have a distinct implementation
-since we should only have to deal with two fibers (one isolate + one main loop
-for its event loop).
+Finally, the isolated context could extend the concurrent context, taking
+advantage of having a single fiber to run —in practice we want to have a
+distinct implementation since we should only have to deal with the thread's main
+fiber and the isolated fiber.
```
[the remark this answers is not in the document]
ysbaddaden: Maybe the current implementation is using the thread's main fiber to run the isolated fiber's block, but after the thread pool it must be a new fiber, because the thread's main fiber is used to park the thread.

text/0002-execution-contexts.md
```diff
@@ -0,0 +1,591 @@
...
+  end
+
+  ncpu.times do
+    codegen.spawn do
+      # (runs in the codegen context)
+      while unit = channel.receive?
+        unit.compile
+      end
+    end
+  end
```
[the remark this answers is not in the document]
ysbaddaden: And here I was wondering why I didn't use `WaitGroup.wait` before :facepalm:

text/0002-execution-contexts.md
```diff
@@ -393,43 +409,47 @@ class SingleThreaded
 end
 ```
 
-A multithreaded context should implement both modules as different types, since
+In practice, the concurrent context might not bring much performance improvement
+over the parallel context with a single thread, and both might share the same
+base.
```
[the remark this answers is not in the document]
straight-shoota: But not on its own, only if you rescale the context. Or am I missing something?

text/0002-execution-contexts.md
```diff
@@ -553,20 +572,20 @@ The **Tokio** crate for **Rust** provides a single-threaded scheduler or a multi
...
+The feature requires both the `preview_mt` and `execution_contexts` flags while we implement and rollout the feature. The first flag is to enable thread safety in stdlib, and the second one to replace the the legacy scheduler with the execution context schedulers.
 
 ~~Since it is a breaking change from stable Crystal, unrelated to `preview_mt`, MT and execution contexts may only become default with a Crystal major release?~~
 
 ~~Maybe MT could be limited to 1 thread by default, and `spawn(same_thread)` be deprecated, but what to do with `spawn(same_thread: true)`?~~
 
 ## Default context configuration
 
-This proposal doesn’t solve the inherent problem of: how can applications configure the default context at runtime (e.g. number of MT schedulers) since we create the context before the application’s main can start. 
+This proposal doesn’t solve the inherent problem of: how can applications configure the default context at runtime (e.g. number of MT schedulers) since we create the context before the application’s main can start.
```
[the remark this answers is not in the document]
straight-shoota: Oh I completely forgot that we already have a PR for rescaling.

straight-shoota (approved):

PR #3, opened 2024-02-06
RFC 0003: WaitGroup

ysbaddaden:
Provide a mechanism to wait on the execution of a set of operations distributed to a set of fibers.

Preview: https://github.com/crystal-lang/rfcs/blob/rfc-wait-group/text/0003-wait-group.md


text/0003-wait-group.md
```diff
@@ -0,0 +1,124 @@
...
+- Mutable counter: a WaitGroup may be modified at any time (but always _before_ the fiber calls `#done`) to increment or decrement the counter. The waiting fibers don't need to know about these changes: they will wait until all the execution is done.
+
+- Signaling fibers: multiple fibers can wait on a WaitGroup, so we can signal a set of fibers at once. For example have a set of fibers wait before starting execution.
+
+# Reference-level explanation 
+
+The proposed API:
+
+```crystal
+class WaitGroup
```
RX14, line 48: Presumably you wouldn't be opposed to helper methods such as 

```cr
def spawn(&)
  add(1)
  begin
    yield
  ensure
    done
  end
end
```

straight-shoota:
Related discussion on the forum: https://forum.crystal-lang.org/t/waitgroup-implementation/4293

mgomes:
Really happy to see this! It feels very Go-centric as is. I think having to manually add fiber counts and decrement via `wg.done` are what make it feel that way.

@RX14's suggested helper method would help a lot. Could the API used by https://github.com/GrottoPress/pond be adapted to be as efficient as this implementation?

It would look like this:

```crystal
def sliced_operation(wg, i)
  32.times do |j|
    wg << spawn do
      sub_sliced_operation(i, j)
    end
  end
end

wg = WaitGroup.new

16.times do |i|
  wg << spawn do
    prepare_slice(i)
    sliced_operation(wg, i)
  end
end

wg.wait
```

text/0003-wait-group.md
```diff
@@ -0,0 +1,124 @@
...
+Go has the sync.WaitGroup type. Java has the CountDownLatch class. Both behave in a similar way as the proposed solution.
+
+The [Earl](https://www.shardbox.org/shards/earl) shard uses a WaitGroup type in its Supervisor and Pool classes to wait on the child fibers it spawned.
+
+# Unresolved questions
+
+What should happen when the counter reached zero? The `#add` and `#done` shall raise a RuntimeError exception, but what about waiting fibers? They might be stuck forever.
+
+- Should the application abort (aka panic)?
+- Should the waiting fibers be resumed _and_ also raise a RuntimeError exception?
```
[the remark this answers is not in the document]
ysbaddaden: - Java only checks the counter at initialization (can't add dynamically), then `countDown()` in OpenJDK is clamped to `0..` which overlooks the issue :shrug: 
- Go panics when `Add` would decrement below 0 (similar to raising an exception), and leaves goroutines waiting (I feel like they won't be resumed if the panic is recovered).

text/0003-wait-group.md
ysbaddaden: That looks nice :+1:

ysbaddaden:
@mgomes Pond has an interesting approach. It's closer to nurseries in structured concurrency (keep a list of child fibers) than a mere counter though, and I feel like it would and up duplicating what such nurseries would do.

Looking at your example, I feel that the `#spawn(&)` helper would do just the same, and I prefer how abstract it is over `#<<(fiber)`.

text/0003-wait-group.md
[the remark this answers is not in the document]
straight-shoota: My take on this: The intention of a wait group is to block waiting fibers until the counter reaches zero.
I think it would be reasonable to still resume the waiting fibers if it overshoots into the negative. It still means we've reached zero on the way...
Of course this is an invalid state, and `#done`/`#add` should raise on going negative.
And the counter should probably reset to 0 as to recover from this state.
Otherwise we would need all further calls to `#wait` raise immediately.

text/0003-wait-group.md
```diff
@@ -0,0 +1,130 @@
...
+# Prior art
+
+Go has the sync.WaitGroup type. Java has the CountDownLatch class. Both behave in a similar way as the proposed solution.
+
+The [Earl](https://www.shardbox.org/shards/earl) shard uses a WaitGroup type in its Supervisor and Pool classes to wait on the child fibers it spawned.
+The [Pond](https://github.com/GrottoPress/pond) shard implements a nursery-like spawner with a waiting mecanism.
+
+# Unresolved questions
+
+What should happen when the counter reached zero? The `#add` and `#done` shall raise a RuntimeError exception, but what about waiting fibers? They might be stuck forever.
```
yxhuvud, line 123: >The `#add` and `#done` shall raise a RuntimeError exception

This I find less than obvious. Consider:

```

16.times do |i|
  spawn do
    executes_fast() 
  ensure
    wg.done
  end
  do_something_that_is_sometimes_slow
end

wg.wait
```

Especially in a multithreaded environment, this will be very racy as the counter may reach 0 any amount of times. It is only once wg.wait is reached that it should be a failure to `#add`. Is the description missing something or some motivation here?

ysbaddaden: @yxhuvud I'm not sure I understand your example. The failure is about reaching a negative number.

We can reach zero multiple times if we call `wg.add(1)` on each iteration (which affects `wg.spawn`). For example:

```crystal
wg = WaitGroup.new
16.times do
  wg.add(1)
  spawn { wg.done }
  do_sometimes { Fiber.yield }
end
wg.wait
```

By the time we end the iteration we'll have incremented the counter 16 times and maybe decremented it up to 16 times already, maybe reaching zero _but_ never a negative number, but we still haven't called `#wait` so it's fine (trying to wait while counter is zero simply returns) —we must be extra careful to avoid race conditions! No exception will be raised.

Now, failing to call `wg.add(1)` or properly setting the initial counter is the error: we _must_ increment _before_ we can decrement:

```crystal
wg = WaitGroup.new(6)
16.times do
  spawn { wg.done } # <= ERROR: negative counter
end
```

**But the fact that the counter may reach zero multiple times is ticking me.** What if we spawn a concurrent fiber that will wait on the waitgroup? Maybe it's a reason why Java's [CountDownLatch](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CountDownLatch.html) doesn't have increments. Though Go's [sync.WaitGroup](https://pkg.go.dev/sync#WaitGroup.Add) has some restrictions, it still allows to increment after there is a waiter (only panics when we try to add while counter is zero and there is a waiter).

```crystal
wg = WaitGroup.new

spawn do
  wg.wait # <= may be resumed early
  do_something
end

16.times do
  wg.add(1)
  spawn { wg.done }
  do_sometimes { Fiber.yield }
end

wg.wait
```

yxhuvud: > The failure is about reaching a negative number.

That may be the intention, but it is definitely not what the text says.

> I'm wondering if we shouldn't prohibit to increment after there is a registered waiter.

One way to do it would be to do something similar to what I do in nested scheduler, ie let `new` take a block. As long as that block is executing it accepts new entries, and once it finishes it does not. In nested_scheduler I enforce the wait after the block has finished executing, but that may potentially not be the best choice for a wait group. In either case, there might be need for more state if there is need to keep track of if it is open for submission or not, and that might not be desirable.

(it also ties in *very* heavily into the lifetime question of the execution contexts - in nested_scheduler every fiber started in the system belongs to a context, so having a complete idea about the lifetime is a necessity)

straight-shoota: We shouldn't overthink it, though. The way I understand it, wait groups are a low level primitive and they'll eventually be used as building blocks for higher level APIs which should cover most typical use cases. As such, wait group needs to be sane and functional but not necessarily eliminate every possible way to shoot yourself in the foot.
People who want to use it directly should read the documentation and be aware of what they're doing. If there's a possibility to end up in an undefined state and that's properly documented, I would be fine with that.

ysbaddaden: > That may be the intention, but it is definitely not what the text says.

Oh my :facepalm:

ysbaddaden: @yxhuvud Then we must synchronize another state (open for increments), and we will or won't be able to increment from a spawned fiber anyway (depends on whether the block has returned or not).

@straight-shoota On one hand I think it's cool to be able to increment at any time, but on the other hand I'm not sure it brings much value over multiple objects, and it might lead to invalid behavior, which is bad. I wouldn't care for an internal object, but this is a public API. It's also easier to lift a restriction than adding one (breaking change).

I'm leaning to add the following restriction: _can't increment once a fiber is waiting_.

straight-shoota: Yeah, my comment was just a general note, not pointing toward anything specific.

straight-shoota: I'm not sure I grasp what invalid behaviour increments can lead to. It should be always safe to increment before any mandatory `done`.

With a wait group construct you always need to pay attention to control flow as you can easily enter an invalid state when the call order of `done`, `add` and `wait` is not as expected.

yxhuvud: > I'm not sure I grasp what invalid behaviour increments can lead to. 

Wakeups before the waitgroup is done, and double wakeups are also a very real possibility. The former breaks the whole promise of waitgroups, whereas the latter just break everything as it isn't possible for the same fiber to execute twice (due to being on the same stack).

ysbaddaden: Argh: if we prevent increments as soon as we have a waiting fiber, then we can't use WaitGroup to implement a dynamic supervisor, or anything dynamic. This means only a static counter would be possible (aka Java's `CountDownLatch`) in which case `#add` makes little sense.

I'd lean to add theses rules to the documentation:

1. must increment before we can decrement;
2. must increment before we can wait;
3. should ideally do all increments before we can wait;
4. must be careful with dynamic increments (that the counter is always at least 1).

Not respecting them will lead to early wakeups.

That being said, **that sounds like _undefined behavior_** and I don't like this :disappointed: 

> double wakeups are also a very real possibility

We can prevent double wakeups (i.e. the PR is safe). I'll state in the RFC that a waiter must be resumed once :+1:

ysbaddaden: I updated the RFC to reflect the issue.

text/0003-wait-group.md
```diff
@@ -118,13 +125,64 @@ Go has the sync.WaitGroup type. Java has the CountDownLatch class. Both behave i
...
+wg = WaitGroup.new
+16.times do
+  wg.add(1)
+  spawn { wg.done }
+  do_sometimes { Fiber.yield }
+end
+wg.wait
+```
+
+By the time the current fiber calls `#wait` we'll have incremented the counter 16 times and decremented it another 16 times; we always increment before we decrement, so we'll never reach a negative counter (that would raise). When the fiber calls `#wait` the counter may be within 0 and 16. If zero the `#wait` method returns immediately, otherwise it suspends the current fiber.
```
straight-shoota, line 144: **question:** Why is this a correctness issue? If we end up with a zero counter at the time `wg.wait` is called, that just means that all things that should've been waited for have already completed, so it should be find for `wait` to just return immediately.

text/0003-wait-group.md
```diff
@@ -118,13 +125,64 @@ Go has the sync.WaitGroup type. Java has the CountDownLatch class. Both behave i
...
+  spawn { wg.done }
+  do_sometimes { Fiber.yield }
+end
+```
+
+The behavior of the loop is identical to the previous example: the counter may reach zero multiple times. The difference is that a concurrent fiber will wait for completion, which is acceptable, yet that fiber is enqueued first, can be resumed at any time and call `#wait` concurrently to the current fiber incrementing the counter. If the counter reaches zero early, the waiting fiber will be resumed early :boom:
+
+We can consider this is breaking the "must increment before we wait" rule, and the proper use is to spawn the waiting fiber after the loop; we could statically set the counter once (`WaitGroup.new(16)`) but then it's not dynamic increment anymore.
+
+We can also consider this as _undefined behavior_ and that can happen under more contrived situations.
```
straight-shoota, line 169: **Question:** I'm wondering if we should call this undefined behaviour. The behaviour of `WaitGroup` is actually well-defined. It's just that the execution of different fibers is non-deterministic. So it may happen that `wg.wait` is called before all `wg.add(1)` have executed and thus the zero counter condition may be reached at a time where not all spawned fibers have completed. That's probably not very useful, but I think it's totally fine from the perspective of the `WaitGroup` API.
If the intention for `do_something_after_completion` is to run after *all* other fibers have completed, this is just a logic error in this code.

text/0003-wait-group.md
```diff
@@ -118,13 +125,64 @@ Go has the sync.WaitGroup type. Java has the CountDownLatch class. Both behave i
...
 
-What should happen when the counter reaches a negative number? The `#add` and `#done` shall raise a RuntimeError exception, but what about waiting fibers? They might be stuck forever.
+**What to do about the undefined behavior explained above?**
+
+- disable dynamic increments (follows Java's CountDownLatch);
+- raise when trying to increment while there is a waiter;
+- pass a block to `#initialize` during which we are allowed to increment;
+- accept the behavior as a known risk and document to be careful (Go seems to do that).
+
+All solutions but accepting the behavior disable dynamic increments in some way, and would prevent using WaitGroup to implement dynamic supervisors or nurseries.
```
straight-shoota, line 180: I'm all for accepting the behaviour as a known risk. This is a low-level concurrency API and it needs to be used responsibly.

text/0003-wait-group.md
```diff
@@ -118,13 +125,64 @@ Go has the sync.WaitGroup type. Java has the CountDownLatch class. Both behave i
...
+We can also consider this as _undefined behavior_ and that can happen under more contrived situations.
+
 # Unresolved questions
 
-What should happen when the counter reaches a negative number? The `#add` and `#done` shall raise a RuntimeError exception, but what about waiting fibers? They might be stuck forever.
+**What to do about the undefined behavior explained above?**
+
+- disable dynamic increments (follows Java's CountDownLatch);
+- raise when trying to increment while there is a waiter;
+- pass a block to `#initialize` during which we are allowed to increment;
```
straight-shoota, line 177: I think this option could even be combined with accpeting the behaviour. If there's ever a need to have a safer API, we could add a yielding overload additionally to the basic, unsafe methods. The idea would be if you use that yielding method, all increments happening in the block would be ensured to be happening before the `wait` call.

text/0003-wait-group.md
ysbaddaden: That one isn't an issue, but a correctness proof.

text/0003-wait-group.md
ysbaddaden: Thanks a lot, that's a very good analysis of where the issue lies: "execution of fibers is non-deterministic" :+1:

beta-ziliani: I guess we can simply delete this line? And maybe expand the previous line to insist on the non-determinism of fibers?

text/0003-wait-group.md
```diff
@@ -0,0 +1,188 @@
...
+- accept the behavior as a known risk and document to be careful (Go seems to do that).
+
+All solutions but accepting the behavior disable dynamic increments in some way, and would prevent using WaitGroup to implement dynamic supervisors or nurseries.
+
+**What should happen when the counter reaches a negative number?**
+
+The `#add` and `#done` raise a RuntimeError exception (this is an error), but the waiting fibers may be stuck forever. Should the waiting fibers be resumed _and_ also raise a RuntimeError exception?
```
beta-ziliani, line 184: > Should the waiting fibers be resumed _and_ also raise a RuntimeError exception?

I think this is better to them being stuck

beta-ziliani (approved):

ysbaddaden:
I resolved the unresolved questions as discussed above.

straight-shoota (approved):

RX14 (approved):

PR #4, opened 2024-02-06
RFC 0004 : HEAP escape analysis

ysbaddaden:
Add an optimization pass to the compiler to detect places where a class instantiation could happen on the stack rather than the GC HEAP.

### DIFF
--- a/text/0004-heap-escape-analysis.md
+++ b/text/0004-heap-escape-analysis.md
@@ -1,0 +1,64 @@
+- Feature Name: `heap_escape_analysis`
+- Start Date: 2024-02-06
+- RFC PR: [crystal-lang/rfcs#4](https://github.com/crystal-lang/rfcs/pull/4)
+- Issue: -
+
+# Summary
+
+Add an optimization pass to the compiler to detect places where a class instantiation could happen on the stack rather than the GC HEAP.
+
+# Motivation
+
+Crystal always allocates class instances into the garbage collected HEAP, even in situations when a variable is only ever accessed locally. This can increase pressure on the GC that needs to allocate then do a collection to reclaim the memory.
+
+Allocating in the HEAP is much slower than allocating on the stack. The GC is also a potential contention point in a multithreaded environment. Having less allocations when possible would improve performance.
+
+# Guide-level explanation
+
+The escape analysis optimization to the semantic pass: if a variable doesn’t outlive the method it is defined in, and its instance size isn’t too large to be allocated on the stack, the compiler can decide allocate it on the HEAP.
+
+A variable can’t outlive a method when the reference:
+
+- isn’t returned by the method;
+- isn’t passed to a method that would capture it (i.e. the argument outlive the method).
+
+The pass automatically decides where to allocate local variables.
+
+To detect more situations more easily, the pass should run after any code inlining passes (e.g. blocks). 
+
+# Reference-level explanation
+
+This is left to actual implementors.
+
+# Drawbacks
+
+Some objects would now be allocated on the stack, which are GC roots that must be scanned conservatively and can't scan precisely. This may have negatively impact GC performance when we implement semi-conservative marking (conservative marking of GC roots _yet_ precise marking of HEAP allocated objects).
+
+# Rationale and alternatives
+
+Developers shouldn't have to worry about where a class is instanciated. The optimization would optimize anything from stdlib, shards or user code, all the while developers would use the objects normally.
+
+Alternatively, developers can already change a `class` into a `struct` but the semantics will change (pass by copy, must pass pointers) is unsafe (no protection against dangling pointers) and limited to the types the developer has actual control upon, as we can't change a type from stdlib or a shard.
+
+There is also pending work to add mechanisms to manually allocate class instances on the stack (see #13481). The advantage is that it gives control to developers, the drawback is that it puts the burden to the developer to use the mechanism, and it can be unsafe because a method may capture a reference (dangling pointer) which Crystal won't protect against.
+
+> [!NOTE]
+The HEAP escape analysis is a best effort and may not detect subtle situations as safe; it may not replace manual stack allocations as decided by the developer.
+
+# Prior art
+
+Java has escape analysis since JAVA SE 6. It seems to be decided at runtime and allows partial HEAP escape (the variable is reallocated to the HEAP on escape).
+
+Go developers don’t decide where variables are allocated: whenever possible Go will allocate variables to the stack, but will allocate them to the GC HEAP if the variable may outlive the function or when the variable is too big to be allocated on the stack.
+
+Wikipedia has an [Escape analysis article](https://en.wikipedia.org/wiki/Escape_analysis), mentioning Java and Scheme.
+
+# Unresolved questions
+
+If the escape analysis is too costly, maybe the optimization may only be enabled when an optimization level is selected? That would avoid negatively impacting development builds.
+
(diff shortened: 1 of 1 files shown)


text/0004-heap-escape-analysis.md
```diff
@@ -0,0 +1,64 @@
...
+- isn’t returned by the method;
+- isn’t passed to a method that would capture it (i.e. the argument outlive the method).
+
+The pass automatically decides where to allocate local variables.
+
+To detect more situations more easily, the pass should run after any code inlining passes (e.g. blocks). 
+
+# Reference-level explanation
+
+This is left to actual implementors.
```
ysbaddaden, line 31: We should detail the algorithms, here:

- For each argument of a method:
  - if the argument is a reference or union with a reference:
    - flag the argument as captured if:
      - the reference is passed to another method that will capture it;
      - the reference is stored in a data structure;
      - the reference is copied as a local variable that will be captured (unless it is _always_ replaced before the capture?).

- For each local variable:
  - if the local variable is a reference or union with a reference:
    - flag the argument as captured if:
      - the reference is returned by the function;
      - the reference is stored in a data structure (ivar, class var, pointer, buffer, ...);
      - the reference is passed to another method that will capture it;
      - the reference is copied as a local variable that will be captured (unless it is _always_ replaced before the capture?).

ysbaddaden: Codegen would then be able to decide where to allocate depending on the "captured" flag of the local variables.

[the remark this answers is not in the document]
ysbaddaden: Good point. We'd even need a begin/ensure block.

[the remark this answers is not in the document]
RX14: Injecting a call to `finalize` should happen in the method where it is allocated, after the last usage of the object. Unfortunately, the object could be used in complicated reference-passing control flow inside the method despite it provably never escaping, so the easiest option would be to call it at the end of the method, right before the stack pointer is decremented. Either that, or tracking every variable which the object could possibly be in and then emitting a call after that. I think the most intuitive time is before the method returns though.

RX14: I think a more precise list could be:

- Passed to a method that captures that argument (transitive property)
- Assigned to a global variable
- Assigned to a class variable
- Assigned to an instance variable
- Returned, unless it was an argument of this method
- `pointerof`
- Passed to an FFI method
  - What about llvm intrinsics?
  - Should we provide annotations for C binding authors to rovide lifetime assertions? It seems dangerous, but powerful.

There may be some I'm missing.

I think the transitive property could be implemented as storing a graph of instantiations' arguments instead of a simple flag. This makes parallel and incremental compilation possible as each method can be calculated independently and then only an extremely short pass to traverse the trees at the end produces the final result.

text/0004-heap-escape-analysis.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Drawbacks
+
+Some objects would now be allocated on the stack, which are GC roots that must be scanned conservatively and can't scan precisely. This may have negatively impact GC performance when we implement semi-conservative marking (conservative marking of GC roots _yet_ precise marking of HEAP allocated objects).
```
RX14, line 35: I think it might be possible to reuse a lot of the escape analysis architecture to implement a shadow stack for each fiber which contains only roots. This could be both precise and a far smaller stack than the actual program stack for the GC to scan. Especially given the stack usage will grow when escape analysis is turned on.

I don't know how that tradeoff compares to the LLVM built in stackmap generator, but it's an idea.

text/0004-heap-escape-analysis.md
straight-shoota: > Should we provide annotations for C binding authors to rovide lifetime assertions? It seems dangerous, but powerful.

Could be an option. But this can be an enhancement further down the line when we discover an actual need for it.
This could often be easily solved with explicit stack allocation.
I suppose the main use case would be boxing arbitrary objects as callback context or similar. Not sure if that's much relevant.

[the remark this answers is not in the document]
straight-shoota: Yeah, and there's absolutely no need to finalize as soon as possible. GC's finalize has no guaranteed time frame either.

text/0004-heap-escape-analysis.md
```diff
@@ -0,0 +1,84 @@
...
+
+As a guideline, the transitive property could be implemented as storing a graph of instantiations' arguments instead of a simple flag. This makes parallel and incremental compilation possible as each method's escape analysis can be calculated independently during typing and then only an extremely short pass to traverse the trees at the end produces the final result.
+
+# Drawbacks
+
+Some objects would now be allocated on the stack, which are GC roots that must be scanned conservatively and can't scan precisely. This may have negatively impact GC performance when we implement semi-conservative marking (conservative marking of GC roots _yet_ precise marking of HEAP allocated objects).
+
+# Rationale and alternatives
+
+Developers shouldn't have to worry about where a class is instanciated. The optimization would optimize anything from stdlib, shards or user code, all the while developers would use the objects normally.
```
yxhuvud, line 57:
```suggestion
Developers shouldn't have to worry about where a class is instantiated. The optimization would optimize anything from stdlib, shards or user code, all the while developers would use the objects normally.
```

text/0004-heap-escape-analysis.md
```diff
@@ -0,0 +1,84 @@
...
+## Lifetime
+
+The following list details the cases when a local variable (including method arguments) is considered captured.
+
+- It is a Reference or an Union with a Reference;
+- It is passed to a method that captures that argument (transitive property);
+- It is assigned to a local variable that will be captured;
+- It is assigned to a global variable;
+- It is assigned to a class variable;
+- It is assigned to an instance variable;
```
straight-shoota, line 33: I suppose these assignments could be invalidated if the variable get a new value assigned during the lifetime of the current scope.

In the following example, the value of `bar` gets assigned to an instance variable, but later the instance variable is reset to its original value. This should not capture bar because when the method exits, its no longer assigned to an instance variable.

```
def foo(bar)
  begin
    old_context, @context = @context, bar
    do_something_with context
  ensure
    @context = old_context
  end
end
```

Not sure if this is getting ahead of things. It shouldn't be necessary for a first implementation, which doesn't need this kind of detailled analyisis. It's probably always possible to refine the algorithm to catch more optimizable cases.

ysbaddaden: But the ivar could be captured in `do_something_with` and it will be hard to notice it. We'd have to track ivars, etc.

ysbaddaden: Even if fully inlined: another thread might capture the ivar.

text/0004-heap-escape-analysis.md
```diff
@@ -0,0 +1,84 @@
+- Feature Name: `heap_escape_analysis`
+- Start Date: 2024-02-06
+- RFC PR: [crystal-lang/rfcs#4](https://github.com/crystal-lang/rfcs/pull/4)
+- Issue: -
+
+# Summary
+
+Add an optimization pass to the compiler to detect places where a class instantiation could happen on the stack rather than the GC HEAP.
```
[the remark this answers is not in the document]
straight-shoota: This article comparing borrow checker / lifetimes handling in Rust and C# gives a short introduction how these languages handle escape analysis: https://em-tg.github.io/csborrow/

PR #5, opened 2024-02-09
RFC 0005: Combinatoric library

meatball133:
Add methods for permutations and combinations for `Int`.

### DIFF
--- a/text/0005-Combinatoric math proposal.md
+++ b/text/0005-Combinatoric math proposal.md
@@ -1,0 +1,142 @@
+- Feature Name: `math-combinatoric-library`
+- Start Date: 2024-02-09
+- RFC PR: [crystal-lang/rfcs#0005](https://github.com/crystal-lang/rfcs/pull/0005)
+- Issue: -
+
+# Summary
+
+Add combinatoric and permutations methods to the standard library which works on integers.
+
+# Motivation
+
+Crystal has an already exsisting methods for doing combinations and permutations but those work on some form of collection which for larger sizes makes them very computanional heavy and high memory usage.
+In the scientific world is there many uses of getting to know the size of combinations or permutations but not the individual combinations or permutations.
+These methods could also have an `Int` implementation which doesnt use any form of collections to reduce on the calculations and memory usage required.
+There are already exsisting libraries which impliments this logic but most of them are unmaintained.
+
+The expected outcome out of this is that Crystal will have a stronger standard library for working with scientific purposes.
+But also when working with just sizes of permutations and combinations will have drasticly faster executions.
+
+# Guide-level explanation
+
+When you just want the size of a combination or permutations so are the `BigInt#combination` or `BigInt#permutation` methods way more efficent when working on arrays.
+They work using a mathematical formula based on factorial of numbers.
+
+To use the methods will you have to include the `BigInt` libaries using `require "big"`.
+Deffine a number then you can use the method on that number with how many combinations you want to use.
+
+```crystal
+require "big"
+
+number : BigInt = 5
+number.combinations(3)
+# => 10
+
+number.combinations(2)
+# => 10
+```
+
+In the examples above can the number be reasoned about having 5 items and wanting to know how many ways those 5 items can be put in groups of 3 or 2. 
+The reason why those 2 examples becomes the same can be reasoned out of 2 perspectives:
+
+- Say you would want to add 5 new features to Crystal but you are only allowed to add 3.
+- That would be the same as resoning that you would want to get groups of 2 features that doesnt get added.
+
+Intrested in learning more: https://en.wikipedia.org/wiki/Combination
+
+Another example:
+
+```crystal
+require "big"
+
+number : BigInt = 20
+number.combinations(2)
+# => 190
+
+number_2 : BigInt = 10
+number_2.combinations(2)
+# => 45
+```
+
+Error handeling examples:
+
+```crystal
+require "big"
+
+number : BigInt = 5
+number.combinations(-4)
+# ArgummentError: combinations can't be done on negativ integers
+
+number2 : BigInt = -5
+number2.combinations(2)
+# ArgummentError: combinations can't be done on negativ integers
+```
+
+As for permutations does that work simiralliy with the use `BigInt` libaries using `require "big"`.
+
+```Crystal
+require "big"
+
+number : BigInt = 5
+number.permutations(4)
+# => 120
+```
+
+Combinations and permutations share a lot with each other.
+The key difference between combinations and permutations is that permutations care about the order.
+Take the last example with new features, say instead you have to make a list of the 3 features you would like to see in the language and the first feature is the one you want to see the most to be implemented.
+Then the order matters you put them in the list.
+
+Permutations doesnt have the same pattern as combinations with `5.combinations(3)` is the same as `5.combinations(2)`.
+Instead for permutations so are `n.combinations(n)` always equal `n.combinations(n - 1)` (as long as n is a posstive number and not zero).
+
+```crystal
+require "big"
+
+number : BigInt = 5
+number.combinations(5)
+# => 120
+
+number.combinations(4)
+# => 120
+```
+
+This is because say you have a list of the 4 features in order you most want to see in the languages.
+Adding one more feature to the list would not add another combinations since there is only 1 feature left of choosing.
+
+# Reference-level explanation
+
+Since both of these methods quickly reach large numbers so do they both have to be implemented under `BigInt`.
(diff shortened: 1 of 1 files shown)


text/0005-Combinatoric math proposal.md
```diff
@@ -0,0 +1,142 @@
...
+# => 120
+```
+
+This is because say you have a list of the 4 features in order you most want to see in the languages.
+Adding one more feature to the list would not add another combinations since there is only 1 feature left of choosing.
+
+# Reference-level explanation
+
+Since both of these methods quickly reach large numbers so do they both have to be implemented under `BigInt`.
+They could either be written manualy using the mathematical formula togehter with `BigInt#factorial` or using a more optimzied version.
```
HertzDevil, line 110: GMP already defines `mpz_bin_ui` for `BigInt#combinations(LibGMP::UI)`. I am not sure whether `#permutations` exists.

text/0005-Combinatoric math proposal.md
```diff
@@ -0,0 +1,142 @@
...
+# Summary
+
+Add combinatoric and permutations methods to the standard library which works on integers.
+
+# Motivation
+
+Crystal has an already exsisting methods for doing combinations and permutations but those work on some form of collection which for larger sizes makes them very computanional heavy and high memory usage.
+In the scientific world is there many uses of getting to know the size of combinations or permutations but not the individual combinations or permutations.
+These methods could also have an `Int` implementation which doesnt use any form of collections to reduce on the calculations and memory usage required.
+There are already exsisting libraries which impliments this logic but most of them are unmaintained.
```
straight-shoota, line 15: Could you add a few references to those libraries? Do some of those already have an optimized implementation?

straight-shoota (commented):
I noticed a couple of spelling and grammar mistakes. That's not too bad, but it noticable hinders the reading effort. Could you perhaps try running a spell checker over the text?
Much appreciated 🙏

text/0005-Combinatoric math proposal.md
```diff
@@ -0,0 +1,142 @@
...
+
+The expected outcome out of this is that Crystal will have a stronger standard library for working with scientific purposes.
+But also when working with just sizes of permutations and combinations will have drasticly faster executions.
```
HertzDevil, line 18: One particularly important use case is preallocating the array used for `Indexable#each_combination`:

```crystal
# note: this should not require `BigInt` to work!
ary = Array(Array(Int32)).new(6.combinations(3) + 1)
[0, 1, 4, 9, 16, 25].each_combination(3) { |v| arr << v }
arr << 0 # okay, no reallocation
```

Whether `Indexable#combinations` should also use the same preallocation is a different question. The standard library is currently not quite consistent in that regard. (Contrast with crystal-lang/crystal#10075)

text/0005-Combinatoric math proposal.md
```diff
@@ -0,0 +1,142 @@
...
+# => 120
+```
+
+Combinations and permutations share a lot with each other.
+The key difference between combinations and permutations is that permutations care about the order.
+Take the last example with new features, say instead you have to make a list of the 3 features you would like to see in the language and the first feature is the one you want to see the most to be implemented.
+Then the order matters you put them in the list.
+
+Permutations doesnt have the same pattern as combinations with `5.combinations(3)` is the same as `5.combinations(2)`.
+Instead for permutations so are `n.combinations(n)` always equal `n.combinations(n - 1)` (as long as n is a posstive number and not zero).
```
HertzDevil, line 91: Typo?

```suggestion
Instead for permutations so are `n.permutations(n)` always equal `n.permutations(n - 1)` (as long as n is a posstive number and not zero).
```

text/0005-Combinatoric math proposal.md
```diff
@@ -0,0 +1,142 @@
...
+
+```crystal
+require "big"
+
+number : BigInt = 5
+number.combinations(5)
+# => 120
+
+number.combinations(4)
+# => 120
```
HertzDevil, line 101:
```suggestion
number.permutations(5)
# => 120

number.permutations(4)
# => 120
```

text/0005-Combinatoric math proposal.md
```diff
@@ -0,0 +1,142 @@
...
+When you just want the size of a combination or permutations so are the `BigInt#combination` or `BigInt#permutation` methods way more efficent when working on arrays.
+They work using a mathematical formula based on factorial of numbers.
+
+To use the methods will you have to include the `BigInt` libaries using `require "big"`.
+Deffine a number then you can use the method on that number with how many combinations you want to use.
+
+```crystal
+require "big"
+
+number : BigInt = 5
```
HertzDevil, line 31: These `TypeDeclaration`s will not compile. They should be `BigInt.new(5)` or `5.to_big_i`

text/0005-Combinatoric math proposal.md
```diff
@@ -0,0 +1,142 @@
...
+number.combinations(-4)
+# ArgummentError: combinations can't be done on negativ integers
+
+number2 : BigInt = -5
+number2.combinations(2)
+# ArgummentError: combinations can't be done on negativ integers
```
HertzDevil, line 72: `#combinations` can be generalized to negative integers by letting `(-n).combinations(k) == (n + k - 1).combinations(k) * (-1) ** k`.

text/0005-Combinatoric math proposal.md
```diff
@@ -0,0 +1,142 @@
...
+
+# Summary
+
+Add combinatoric and permutations methods to the standard library which works on integers.
+
+# Motivation
+
+Crystal has an already exsisting methods for doing combinations and permutations but those work on some form of collection which for larger sizes makes them very computanional heavy and high memory usage.
+In the scientific world is there many uses of getting to know the size of combinations or permutations but not the individual combinations or permutations.
+These methods could also have an `Int` implementation which doesnt use any form of collections to reduce on the calculations and memory usage required.
```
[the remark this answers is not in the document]
meatball133: Yeah but combinatoric calculations work on very big numbers. I don't know if a more efficient method can be found. but the one which is defined in math. For example `20.combinations(5)`. Would do `20!`, which is already 2.432902e+18. And then divide that number, you could likely optimize it a bit, but it is still big numbers which are being worked on. So aslong as the group is small it is fine but when it passes like 50 for even an optimized version so do it require bigint.

text/0005-Combinatoric math proposal.md
meatball133: Well you are sorta never dealing with negative combinations, and this method in Python and my calculator returns error when trying to use a negative integer. Then also the combination method is in math derived from factorial and in Crystal does the factorial method not accept negative integers.

HertzDevil: But you said that these methods contribute to "a stronger standard library for working with scientific purposes", and mathematical science dictates that this generalization is valid, otherwise there wouldn't be a reason GMP explicitly states that [negative `n` is supported](https://gmplib.org/manual/Number-Theoretic-Functions) ([original Knuth paper](https://doc.lagout.org/science/0_Computer%20Science/2_Algorithms/The%20Art%20of%20Computer%20Programming%20(vol.%201_%20Fundamental%20Algorithms)%20(3rd%20ed.)%20%5BKnuth%201997-07-17%5D.pdf)).

IMO the restriction to non-negative `n` here (even `k`) is too artificial.

text/0005-Combinatoric math proposal.md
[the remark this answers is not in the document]
HertzDevil: A good start is something like `20 // 1 * 19 // 2 * 18 // 3 * 17 // 4 * 16 // 5`. The divisions are all exact, because the first `n` multiplied values must have exactly one multiple of `n` itself, and this will minimize the possibility of intermediate overflows. It makes sense that `50.combinations(2)` would simply do `50 * 49 // 2` without ever calling any factorial function.

text/0005-Combinatoric math proposal.md
meatball133: I mean, I am perhaps not the best at math, but I find the problem with negative numbers is that most of this works around vectors. Vectors' size can't be negative, but their direction can be. This means if we have the vector : {3, 4} and would multiply it by -1, then it would be {-3, -4}. Its direction is different, but its length or size remains the same, which would be: (-3 ^ 2 + -5 ^2) ^ 0.5 = 5. If we start allowing negative integers, then we start claiming that these vectors' size is negative, which is mathematically impossible without using complex numbers.

With this said, I have a hard time seeing how this can be generalized with negative numbers. I barely know anything about GMP, and I am up for changes if it is possible to generalize these methods, but I am not aware of how that can be implemented.

piero1100:
Kinds of crystallography lances

PR #7, opened 2024-05-02
RFC 0007: Event loop refactor

straight-shoota:
This RFC is very much work in progress and has a lot of unresolved questions which I intend to discuss here.

Preview: https://github.com/crystal-lang/rfcs/blob/rfc/0005/text/0007-event_loop-refactor.md


text/0007-even_loop-refactor.md
```diff
@@ -0,0 +1,168 @@
...
+
+- DNS resolution? ([Async DNS resolution #13619](https://github.com/crystal-lang/crystal/issues/13619))
+- OS signals?
+- File system events?
+- OS event (eventfd) ?
+- User events
+
+Should events from the Crystal runtime be part of the event loop as well?
+
+- Fiber: sleep
```
yxhuvud, line 93: Yes, under the assumption that it is sleep with argument and not endless sleep. At least for io_uring the easiest implementation is to just emit the appropriate op (ie `timeout`) and it will just work in a way that is extremely [straightforward](https://github.com/yxhuvud/nested_scheduler_io_uring_context/blob/main/src/nested_scheduler/io_uring_context.cr#L235).

text/0007-even_loop-refactor.md
```diff
@@ -0,0 +1,168 @@
...
+- **System selector:** The system implementation of an event loop which the runtime component depends on to receive events from the operating system. Example implementations: `epoll`, `kqueue`, `IOCP`, `libevent` (a cross-platform wrapper of the former), `io_uring`
+
+- **Scheduler:** manages fibers’ executions inside the program (controlled by Crystal), unlike threads that are scheduled by the operating system (outside of the accessibility of the program).
+
+## General Design Principles
+
+- **Generic API**: Independent of a specific event loop design
+- **Black Box**: No interference with internals of the event loop
+- **Pluggable**: It must be possible to compile multiple event loop implementations and choose at runtime.
+  Only one type of implementation will typically be active (it would be feasible to have different event loop implementations in different execution contexts, if there's a use case for that).
```
yxhuvud, line 45: (Feasible, but problematic). Some default values may be wanted to be different when using file descriptors in different event loops, like for example if sockets are opened in nonblocking mode or not.

text/0007-even_loop-refactor.md
```diff
@@ -0,0 +1,168 @@
...
+- User events
+
+Should events from the Crystal runtime be part of the event loop as well?
+
+- Fiber: sleep
+- Select actions: send, receive, timeout
+
+### Optional event loop features
+
+Some activities are managed on the event loop on one platform but not on others. Example would be `Process#wait` which goes through IOCP on Windows but on Unix it’s part of signal handling. (Note: Perhaps we could try to get that on the event loop on Unix as well? **🤔** But there are other examples of system differences)
```
yxhuvud, line 98: linux has `signalfd` which could help, but I am not aware if mac/bsd have any good solutions.

text/0007-even_loop-refactor.md
```diff
@@ -0,0 +1,168 @@
...
+Should we put them entirely into the scheduler / event loop driver? We’re already doing that in IOCP. Managing these events entirely in Crystal would allow us to drop `Crystal::System::Event` which is only really needed for libevent.
+
+### `#connect` timeout
+
+`Socket#connect` is the only method with a timeout parameter. This seems weird. All other timeouts are configured on the IO instance.\
+It probably makes sense to standardize, introduce a new property \`Socket#connect\_timeout\`, and deprecate the def parameter.
+
+### Bulk events without fibers
+
+For some applications it might be useful to interact with the event loop directly, being able to push operations in bulk without having to spawn (and wait) a fiber for each one.
```
yxhuvud, line 149: One could also think of the reverse. For example io_uring supports multishot `accept`, ie it is not necessary to rearm the op after getting one without emitting more events, so a usage of it could either spawn new fibers or reuse a set of existing fibers with each trigger.

text/0007-even_loop-refactor.md
ysbaddaden: BSD and Darwin have `kqueue` that can receive signals (and many other things).
Recent Linux kernels have `pidfd` that look even more lovely than `signalfd`.

straight-shoota: It's good to know we have the option to put signals on the event loop. It's a different question whether we'd want that. It *could* be useful as it would help simplify the implementation, I imagine?

But this is just an example. The real question here is how to handle events that need to be on the event loop on one system and we cannot put them on on another one. Not sure if there's any currently relevant use case except signals, but it's something to ponder for future extensibility.

text/0007-even_loop-refactor.md
Blacksmoke16, line 1: Seems there's a typo in the file name ^

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,252 @@
...
+
+  # Accepts an incoming TCP connection on the socket and continues fiber when a
+  # connection is available.
+  # Returns a handle to the socket for the new connection.
+  abstract def accept(socket : ::Socket) : ::Socket::Handle?
+
+  # Opens a connection on *socket* to the target *address* and continues fiber
+  # when the connection has been established.
+  # Returns `IO::Error` but does not raise.
+  abstract def connect(socket : ::Socket, address : ::Socket::Addrinfo | ::Socket::Address, timeout : ::Time::Span?) : IO::Error?
```
straight-shoota, line 107: The `address` type could perhaps be abstracted. We need to support both types, though they both boil down to the same interface: `#to_unsafe` returning a pointer to a `LibC::Sockaddr` and `#size` returning its size in bytes.

So it could be an option to define an interface for that and use that as type restriction. But we cannot actually make it strictly type-safe because the return type of `#to_unsafe` is `Void*` and we cannot enforce a pointer to `LibC::Sockaddr` (so `Array` would implement this interface for example).

Alternatively, we could remove the type restriction entirely because any type that fulfills the implicit interface should work.

I think it's better to be explicit, either with `::Socket::Addrinfo | ::Socket::Address` or an abstract module interface.
Introducing a new model would also solve the issue that the `Socket` name space is not in core lib (see https://github.com/crystal-lang/rfcs/blob/rfc/0005/text/0007-event_loop-refactor.md#socket).

dsisnero:
Here's a good link on CancelToken instead of looking for Timeouts or even signals - use a CancelToken and have those as subclasses of it https://vorpus.org/blog/timeouts-and-cancellation-for-humans/

text/0007-even_loop-refactor.md
straight-shoota: To properly solve this, I suppose it would require the opening of file descriptors to be passed to the event loop as well.

yxhuvud: Deferring it to the event loop help a little but doesn't really solve it as it means fds cannot be passed around. So for example, it would not work to open the socket in one context, and then start to listen on it and spawn handlers in another. And if descriptors are not setup properly, the error symptoms would be weird. Not having that confusion and that footgun around is worth more than what it would fix.

My preference for linux in the really long term would be to keep the default to nonblocking until there is no longer any support for any linux version that doesn't have a sufficiently modern io_uring, and then switch the default once we have an implementation we are happy with. The latter can be implemented using poll ops until the switch (similar to the uring branch) - less neat and does more work than necessary, but it would work.  But at some time it would be nice to switch as the polling adds a little overhead both to code and performance. 

 (FWIW, note that there is still cases where `wait_readable` and `wait_writable` is useful even with io_uring. perhaps their fate needs to be a separate discussion as they don't provide value/make sense on certain other platforms)

straight-shoota: Yeah, I don't think it makes sense to move open fds between different event loop implementations. That's just going to be chaos. And I'm not sure it would even be very useful anyway.
However I could potentially see use cases for using one kind of event loop implementation for a part of an application, and another one in another part. Not saying that this definitely makes sense, but it might.

text/0007-event_loop-refactor.md
straight-shoota: It's actually hard to stub out `::Socket::Addrinfo` and `::Socket::Address` because that would require defining the `Socket` namespace, which is a class that inherits `IO` and that's so much complexity we could just include `socket.cr`.

I've considered introducing a module which could be included in both types. But that's actually not a great solution for when we return an address.
So I think it's probably best to introduce an internal struct type which holds a reference to `LibC::Sockaddr` and its size.

text/0007-even_loop-refactor.md
yxhuvud: > Yeah, I don't think it makes sense to move open fds between different event loop implementations.

So what you are saying is that each event loop need to open their own copy of std_* and keep track of them separately? And that it shouldn't be possible to safely pass file descriptors around without restrictions? That seems a lot more chaotic to me. The gains for adding restrictions like that on the user seems very tiny. 

At least on linux/mac. No idea what would be the best choice on windows.

straight-shoota: As a use case I was imagining a set of file descriptors being handled by a separate event loop from the rest of the application. This could be a set of sockets for some specific communication purpose. There doesn't need to be any standard IO on such an event loop. Or any other interference with file descriptors that are not part of that set.

ysbaddaden: If I understand correctly we want to have multiple event loop implementations compiled together (i.e. have io_uring + epoll on Linux) but only one implementation will be activated after a runtime check for the whole program.

Couldn't the event loop implementation tell whether it wants O_NONBLOCK for example?

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,252 @@
...
+
+#### Socket
+
+One instance of this problem shows already in the core features: The event loop interface has type restrictions of the `Socket` namespace in abstract defs, but `Socket` is not in the core lib.
+
+Options:
+
+- Omit those abstract defs (dilutes the interface, so not ideal)
+- Split `EventLoop` interface and add parts of it only with `require “socket”`
+- Add stub declarations for the involved types (`Socket::Handle` and `Socket::Address` - `Socket` itself is only used as parameter type which is technically okay for abstract methods)
```
straight-shoota, line 210: I considered option 3 (stubbing the affected types in core lib) as maybe the most approachable solution, but I don't think it is.
We technically don't need to stub `Socket` because it's only used as type restriction in the parameters which are not validated by the abstract method checker. However, I believe this should eventually be covered as well, which would break this solution in the future. And I think it's generally better to be strict about abstract method implementations, even if the compiler won't complain.
Either way, I don't think it makes much sense to stub that many types for the event loop API. It would probably be easier to just use the actual types directly.

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,252 @@
...
```
straight-shoota, line 194: I considered option 3 (stubbing the affected types in core lib) as maybe the most approachable solution, but I don't think it is.
We technically don't need to stub `Socket` because it's only used as type restriction in the parameters which are not validated by the abstract method checker. However, I believe this should eventually be covered as well, which would break this solution in the future. And I think it's generally better to be strict about abstract method implementations, even if the compiler won't complain.
Either way, I don't think it makes much sense to stub that many types for the event loop API. It would probably be easier to just use the actual types directly.

straight-shoota: Considering that we may need to handle optional event loop features eventually (see previous section), it might be a good exercise to start with sockets.

Here's how this could work:
* Define a module `Crystal::System::EventLoop::Socket`
* This module is empty by default.
* In `crystal/system/socket.cr`, add abstract defs to `Crystal::System::EventLoop::Socket` (this could probably be in a different file).
* All current event loop implementations implement the socket interface, so they can unconditionally include `Crystal::System::EventLoop::Socket`. We can even include it in `Crystal::System::EventLoop` (and pull it out when we ever get an event loop implementation that *doesn't* implement sockets).
* In a program where sockets are not used, the abstract defs are not checked because the module is empty. The implementation methods are never called, so their type restrictions don't matter.
* In a program where sockets are used, the module has abstract defs and their implementation will be checked.

```cr
module Crystal::System::EventLoop::Socket
end

abstract class Crystal::System::EventLoop
  include Socket
end
```
```cr
# src/crystal/system/socket.cr

module Crystal::System::EventLoop::Socket
  abstract def close(socket : ::Socket) : Nil

  # etc.
end
```

```cr
# src/crystal/system/unix/event_loop_libevent.cr
class Crystal::LibEvent::EventLoop
  def close(socket : ::Socket) : Nil
    # ...
  end

  # etc.
end
```

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,252 @@
...
+
+  # Writes at least one byte from *slice* to the socket with a target *address* (UDP)
+  # and continues fiber when the write is complete.
+  # Returns the number of bytes written.
+  abstract def send_to(socket : ::Socket, slice : Bytes, address : ::Socket::Address) : Int32
+
+  # Receives on the socket into *slice*  and continues fiber when the package is
+  # completed
+  # Returns the number of bytes received.
+  abstract def receive(socket : ::Socket, slice : Bytes) : Int32
```
straight-shoota, line 122: I'm wondering if this variant is even necessary. It would currently not be used by the stdlib socket implementation. All public API use cases in variants of `Socket#receive` include the source address.

text/0007-event_loop-refactor.md
beta-ziliani: > So I think it's probably best to introduce an internal struct type which holds a reference to LibC::Sockaddr and its size.

Or to leave it as the union?

text/0007-event_loop-refactor.md
beta-ziliani: If the implementation won't use it, then I agree there's no point in having it. Or to put it otherwise, what would be a reason to have it?

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,252 @@
...
+- **Black Box**: No interference with internals of the event loop
+- **Pluggable**: It must be possible to compile multiple event loop implementations and choose at runtime.
+  Only one type of implementation will typically be active (it would be feasible to have different event loop implementations in different execution contexts, if there's a use case for that).
+
+# Reference-level explanation
+
+The new `EventLoop` defines the following interface for issuing operations on the event loop:
+
+```cr
+module EventLoop
```
beta-ziliani, line 78: It'd be nice to have a homogeneous handling of errors: Would it make sense to let all functions return an error?

text/0007-event_loop-refactor.md
beta-ziliani: I know I was an advocate for this style, but now that I see it I'm having a little itch. How would it look like with a split interface for the EventLoop? Say, `EventLoop::SocketInterface`, `EventLoop::FileInterface`, ... and then let each EL implementation include the module interfaces they comply to.

text/0007-event_loop-refactor.md
straight-shoota: Those types are not in core lib. Their existence depends on `require "socket"`. So leaving the union won't work as is. Trying to see if we can stub it out (or replace with an internal type) is one possible way to deal with that.

But I think a better approach is described in https://github.com/crystal-lang/rfcs/pull/7#discussion_r1615841842

text/0007-event_loop-refactor.md
straight-shoota: Yeah, that's basically what this is.
However, there's the problem that at the event loop implementation we can't know whether sockets are used in the program (i.e. if `Socket` is defined) or not. The event loop gets required as part of the prelude while `require "socket"` will only be in user code.

So the trick here is to have the socket interface empty by default and only fill it with abstract defs if sockets are used.
That makes it possible for the event loop implementation to always include the socket interface. We get abstract def checks if sockets are used.

text/0007-event_loop-refactor.md
straight-shoota: It avoids the memory allocation and initialization of the source address if that's irrelevant.
However, `#receive` should actually be equivalent to `#read`, so you can just use that.

straight-shoota: Oh, this makes we realize, we actually don't need both, `write` and `send`, either. They should always be equivalent.

Their existing implementations are identical.

* Win32: [Socket#system_send](https://github.com/crystal-lang/crystal/blob/a30e3d96248a55de4524005837f944294dd14278/src/crystal/system/win32/socket.cr#L262-L271), [Socket#system_write](https://github.com/crystal-lang/crystal/blob/a30e3d96248a55de4524005837f944294dd14278/src/crystal/system/win32/socket.cr#L460-L469)
* Unix: [Socket#system_send](https://github.com/crystal-lang/crystal/blob/a30e3d96248a55de4524005837f944294dd14278/src/crystal/system/unix/socket.cr#L82-L86), [Socket#system_write](https://github.com/crystal-lang/crystal/blob/a30e3d96248a55de4524005837f944294dd14278/src/crystal/system/unix/socket.cr#L259-L263).

text/0007-event_loop-refactor.md
straight-shoota: Not sure I follow. Do you mean we should replace the error mechanism in all code used by the event loop implementations to return an error value instead of raising an exception?

straight-shoota: The reason why `connect` returns errors is an optimization to avoid the cost of raising an exception. When the connection address is an `Addrinfo` there could be multiple possible values and the socket will try connecting to each one, stopping as soon as it is successful.
That's the only place where an error needs to be handled internally, so we can avoid the overhead. In all other event loop actions this is not necessary, so we just raise directly (instead of in the code calling into the event loop). Changing that just for the sole purpose of standardizing the event loop interface seems unreasonable. Then I'd rather look into whether we can drop the optimization and always raise (and rescue the connect call in `TCPSocket#initialize`).

beta-ziliani: Consistency was part of it, true, but not the most relevant one. WASI didn't (doesn't?) have exceptions (well, or anything else :-D), so having an alternate stdlib with no exceptions make sense in such smaller environments, and I dream of an implementation where at the core are very basic blocks...

But it was just a suggestion, if it makes the next steps harder, then there's no point.

straight-shoota: The event loop implementation uses the standard library, so we would first need to start with the basic parts of stdlib before making the event loop exception-free.

text/0007-event_loop-refactor.md
straight-shoota: I created a PR to drop `system_send`: https://github.com/crystal-lang/crystal/pull/14637

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,245 @@
...
+  module Socket
+    # Reads at least one byte from the socket into *slice* and continues fiber
+    # when the read is complete.
+    # Returns the number of bytes read.
+    abstract def read(socket : ::Socket, slice : Bytes) : Int32
+
+    # Writes at least one byte from *slice* to the socket and continues fiber
+    # when the write is complete.
+    # Returns the number of bytes written.
+    abstract def write(socket : ::Socket, slice : Bytes) : Int32
```
straight-shoota, line 107: I'm wondering if we should use the terms `read`/`write` or `send`/`receive`. Their meaning is essentially equivalent.
`read`/`write` would be more similar to the `FileDescriptor` equivalents.
But the imlementations tend to use the `send`/`receive` system APIs. So I might prefer that.

straight-shoota:
I have created two PRs implementing the base event loop interfaces:

* `FileDescriptor`: https://github.com/crystal-lang/crystal/pull/14639
* `Socket`: https://github.com/crystal-lang/crystal/pull/14643

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,295 @@
...
+    abstract def read(socket : ::Socket, slice : Bytes) : Int32
+
+    # Writes at least one byte from *slice* to the socket.
+    #
+    # Blocks the current fiber if the socket is not ready for writing,
+    # continuing when ready. Otherwise returns immediately.
+    #
+    # Returns the number of bytes written (up to `slice.size`).
+    #
+    # Use `#receive_from` for capturing the source address of a message.
```
beta-ziliani, line 142: These lines got scrambled?

```suggestion
    # Use `#receive_from` for capturing the source address of a message.
    abstract def read(socket : ::Socket, slice : Bytes) : Int32

    # Writes at least one byte from *slice* to the socket.
    #
    # Blocks the current fiber if the socket is not ready for writing,
    # continuing when ready. Otherwise returns immediately.
    #
    # Returns the number of bytes written (up to `slice.size`).
    #
    # Use `#send_to` for sending a message to a specific target address.
```

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,295 @@
...
+  #       time in parallel, but this assumption may change in the future!
+  abstract def interrupt : Nil
+
+  include FileDescriptor
+  include Socket
+
+  module FileDescriptor
+    # Reads at least one byte from the file descriptor into *slice* and continues
+    # fiber when the read is complete.
+    # Returns the number of bytes read.
```
ysbaddaden, line 111: As suggested in the PR, let's be overly descriptive of the behavior:

```suggestion
    # Reads at least one byte from the file descriptor into *slice*.
    #
    # Blocks the current fiber if no data is available for reading, continuing when available. Otherwise returns immediately.
    #
    # Returns the number of bytes read (up to *slice.size*).
    # Returns 0 when EOF is reached.
```

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,295 @@
...
+    abstract def read(file_descriptor : Crystal::System::FileDescriptor, slice : Bytes) : Int32
+
+    # Writes at least one byte from *slice* to the file descriptor and continues
+    # fiber when the write is complete.
+    # Returns the number of bytes written.
```
ysbaddaden, line 116:
```suggestion
    # Writes at least one byte from *slice* to the file descriptor.
    #
    # Blocks the current fiber if the file descriptor isn't ready for writing, continuing when ready. Otherwise returns immediately.
    #
    # Returns the number of bytes written (up to *slice.size*).
```

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,295 @@
...
+    # becomes available. Otherwise returns immediately.
+    #
+    # Returns a handle to the socket for the new connection.
+    abstract def accept(socket : ::Socket) : ::Socket::Handle?
+
+    # Opens a connection on *socket* to the target *address*.
+    #
+    # Blocks the current fiber and continues when the connection is established.
+    #
+    # Returns `IO::Error` in case of en error. The caller is responsible for
```
ysbaddaden, line 157:
```suggestion
    # Returns `IO::Error` in case of an error. The caller is responsible for
```

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,295 @@
...
+    # continuing when ready. Otherwise returns immediately.
+    #
+    # Returns the number of bytes written (up to `slice.size`).
+    #
+    # Use `#send_to` for sending a message to a specific target address.
+    abstract def write(socket : ::Socket, slice : Bytes) : Int32
+
+    # Accepts an incoming TCP connection on the socket.
+    #
+    # Blocks the current fiber if no connection is watiting, continuing when one
```
ysbaddaden, line 147:
```suggestion
    # Blocks the current fiber if no connection is waiting, continuing when one
```

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,295 @@
...
+    # Blocks the current fiber if the socket is not ready for writing,
+    # continuing when ready. Otherwise returns immediately.
+    #
+    # Returns the number of bytes sent (up to `slice.size`).
+    abstract def send_to(socket : ::Socket, slice : Bytes, address : ::Socket::Address) : Int32
+
+    # Receives at least one byte from the socket into *slice*, capturing the
+    # source address.
+    #
+    # Blocks the current fiber if if no data is available for reading, continuing
```
ysbaddaden, line 173:
```suggestion
    # Blocks the current fiber if no data is available for reading, continuing
```

ysbaddaden (commented):
I left some suggestions for the API documentation (typos, improve FD#read and FD#write).

yxhuvud:
Based on comment in the [read_evented](https://github.com/crystal-lang/crystal/issues/14747#issuecomment-2189649843 )  issue, I'm noting that support for `wait_readable` (and probably `wait_writable`, though that one is less used) would be nice to provide on the systems and implementations where it is possible. It is a fairly common ideom in C libraries on Linux, and it ties in extremely well with an event loop that is built on epoll, kevent or io_uring. Examples of usage: [v4l2.cr](https://github.com/postmodern/v4l2.cr), [wayland_client](https://github.com/yxhuvud/wayland_client). I don't know enough about the former to say if it actually needs to wait on readability or if the fixes in 1.12 was sufficient for that use case, but the latter would have to be rewritten from scratch without using system libraries if the ability to wait on readability wasn't there.

As these cannot be implemented on windows, but is at the same time very convenient to have access to where it is possible, I wonder if there should be something written about them in the rfc.

ysbaddaden:
Damn, those are valid use cases.

straight-shoota:
Obviously the event loop implementations for libevent, kqueue, epoll, io_uring etc. will always require something like `wait_readable`/`wait_writable` as building blocks. So that functionality must be available on supporting platforms.

But it probably fits best as a feature of the event loop implementation, not the file descriptor model. The former is platform-specific, while the latter is (supposed to be) platform-independent.
It would then be possible to directly interface with the event loop to implement the features that you mention.

ysbaddaden:
@straight-shoota that's what I was thinking: an optional interface for the event loop to implement? Windows would raise NotImplementedError (until we can figure something, like running a WSAPoll in its own thread?)

ysbaddaden:
After implementing the Epoll (and Kqueue) event loops:

1. I wish `sleep` was a mere `Crystal::EventLoop.sleep(duration : Time::Span)`. The implementation would be much simpler and direct: create a timer event on the stack, suspend the fiber, done.

2. `Crystal::EventLoop::Event` is missing a `#delete : Nil` method (called when cancelling a select timeout).

straight-shoota:
I have updated the RFC text to reflect the implemented status. So I think it's ready to be merged.

There are a number of *Unresolved questions* but I don't think they need to be resolved in the scope of this RFC.
We need to figure out what to do with them. Some might turn into their own actionable items. Others may just stay here for reference.

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,308 @@
...
+Should events from the Crystal runtime be part of the event loop as well?
+
+- Fiber: sleep
+- Select actions: send, receive, timeout
+
+### Optional event loop features
+
+Some activities are managed on the event loop on one platform but not on others. Example would be `Process#wait` which goes through IOCP on Windows but on Unix it’s part of signal handling. (Note: Perhaps we could try to get that on the event loop on Unix as well? **🤔** But there are other examples of system differences)
+
+Do we require these optional methods to be present in all event loop implementations, i.e. they’re part of the global interface? Some impls would then just raise “Not implemented”. Alternatively, we could keep them out of the main interface and check for availability via `event_loop.responds_to?`. or a sub interface (`is_a?(EventLoop::Process)`). Or…?
```
beta-ziliani, line 265: We have an answer to these now, we should specify what we did

beta-ziliani (commented):
The last bit can be refined now to explain or at least comment what we ended up doing

beta-ziliani:
> There are a number of Unresolved questions but I don't think they need to be resolved in the scope of this RFC.
We need to figure out what to do with them. Some might turn into their own actionable items. Others may just stay here for reference.

We can place them under "future possibilities". In particular, the `select` part, which likely requires its own rfc

ysbaddaden (approved):

text/0007-event_loop-refactor.md
```diff
@@ -134,7 +144,7 @@ abstract class Crystal::System::EventLoop
     # Returns the number of bytes read (up to `slice.size`).
     # Returns 0 when the socket is closed and no data available.
     #
-    # Use `#receive_from` for capturing the source address of a message.
+    # Use `#send_to` for sending a message to a specific target address.
```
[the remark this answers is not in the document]
straight-shoota: The text is copied exactly from the stdlib implementation. Same error there🤦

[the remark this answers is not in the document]
straight-shoota: https://github.com/crystal-lang/crystal/pull/15194

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,308 @@
...
+
+## References
+
+- <https://tinyclouds.org/iocp_links>
+- Zig: [Proposal: Event loop redesign (ziglang/zig#8224)](https://github.com/ziglang/zig/issues/8224)
+- [\[RFC\] Fiber preemption, blocking calls and other concurrency issues (crystal-lang/crystal#1454)](https://github.com/crystal-lang/crystal/issues/1454)
+
+# Unresolved questions
+
+- Distinction between generic system APIs and event loop-specifics is not always clear.
```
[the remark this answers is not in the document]
straight-shoota: No I suppose we can drop it.

text/0007-event_loop-refactor.md
straight-shoota: Do we have a clear answer? 🤔 I don't think we have really worried about this because it hasn't come up yet. All event loop implementations per platform cover the same scope. Even the lifetime event loops have exactly the same operations as the libevent implementation.
This question would become more relevant when we implement for example pidf for the epoll driver, but not for libevent.

beta-ziliani: Perhaps a clarification that this is for the future?

text/0007-event_loop-refactor.md
```diff
@@ -0,0 +1,308 @@
...
+* `Socket` is not part of the core library, so references to its types (e.g. in type restrictions) would not resolve.
+  The `EventLoop` interface is split into individual modules and the `Socket` module is always defined, but only filled with abstract defs when `socket.cr` is explicitly required.
+
+# Drawbacks
+
+None.
+
+# Rationale and alternatives
+
+TBD
```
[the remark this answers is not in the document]
beta-ziliani: I thought this was obvious, but there's no point in having a section with a TBD. Either we fill it or we drop it entirely. Maybe ask ChatGPT? (kidding!)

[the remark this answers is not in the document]
beta-ziliani: Alternatively, we can link to Motivation (Rationale) and Future possibilities (alterantives)

text/0007-event_loop-refactor.md
straight-shoota: It's under *Future possibilities* now?

text/0007-event_loop-refactor.md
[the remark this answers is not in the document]
ysbaddaden: We can drop the section.

The rest of the RFC already explains the rationale (we must support different backends that are system dependent) and we don't need to fathom alternatives just because.

Or maybe just a simple:

```suggestion
The rationale is to create our own system independent event loop, and to keep it as abstract as possible to support any kinds of models (asynchronous, polling) over the actual operations.

Alternatives, as they exist today, leak system specifics into the stdlib types (see `IO::Evented`).

The proposed model completely eliminates that by creating a standardized interface.
```

beta-ziliani (approved):

PR #11, opened 2024-11-14
RFC 0011: Extending API Docs

nobodywasishere:
This RFC is a follow up to discussion on the [Crystal forum](https://forum.crystal-lang.org/t/rfc-extending-api-docs-and-compiler-semantic-output/7314).

### DIFF
--- a/text/0011-extending-api-docs.md
+++ b/text/0011-extending-api-docs.md
@@ -1,7 +1,9 @@
-- Feature Name: extending-api-docs
-- Start Date: 2024-11-14
-- RFC PR: [crystal-lang/rfcs#11](https://github.com/crystal-lang/rfcs/pull/11)
-- Issue: [crystal-lang/crystal#6721](https://github.com/crystal-lang/crystal/issues/6721)
+---
+Feature Name: extending-api-docs
+Start Date: 2024-11-14
+RFC PR: "https://github.com/crystal-lang/rfcs/pull/11"
+Issue: "https://github.com/crystal-lang/crystal/issues/6721"
+---
 
 # Summary
 
--- a/text/0011-extending-api-docs.md
+++ b/text/0011-extending-api-docs.md
@@ -20,7 +20,7 @@ into the source code in order to figure out what methods are available.
 
 # Guide-level explanation
 
-The `:showdoc:` directive can be added to private or protected objects, as well as C lib binding objects, to have them show up in API documentation.
+The `:showdoc:` directive can be added to private or protected objects, methods, and C lib bindings, to have them show up in API documentation.
 By default, these are hidden and should only be shown if they're explicitly intended to be used.
 
 In this example, when generating API documentation, `Foo.foo` will be included even though it is a private method.
--- a/text/0011-extending-api-docs.md
+++ b/text/0011-extending-api-docs.md
@@ -76,16 +76,20 @@ end
 - The parser will need to be updated to support doc comments for C lib binding objects and the `:showdoc:` directive
 - The documentation generator will need to be updated to support C lib binding objects and private/protected objects
 - If an object has a `:showdoc:` directive and its parent namespace is shown, then it should be shown too
-
-TBD
+- If a C lib has a `:showdoc:` directive, everything in the namespace should be shown, except for things marked with `:nodoc:`
+- If an objects parent namespace has the `:nodoc:` directive, the `:showdoc:` directive will have no effect
+- C functions should include both the original function name and the Crystal name in the documentation
+- Some experimentation into a different implementation was done [here](https://github.com/crystal-lang/crystal/compare/master...nobodywasishere:crystal:nobody/docs-include-more)
 
 # Drawbacks
 
-TBD
+Allowing showing private or protected methods in documentation may lead to confusion from users when they try to use those methods (and they don't work).
 
 # Rationale and alternatives
 
 The other design that has been considered is having flags on the documentation generator itself that enable showing of private / protected objects and C lib objects in the API documentation. We chose not to go with this design as it required flags to be added at generation time, and only generated all or none (no granularity in what is shown).
+
+Another alternative is a more sophisticated syntax for documentation metadata, like [sections](https://github.com/crystal-lang/crystal/issues/1312). This would require more complex documentation comment parsing but would allow for more complex and sophisticated documentation features. Inspiration could be taken from [rST field lists](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#field-lists). The `:nodoc:` and `:showdoc:` directives could've been implemented as annotations instead, but this wouldn't allow for features like sections.
 
 This cannot be done in a library instead as it requires updates to the parser itself. This proposal makes Crystal code easier to understand, as it increases the amount and quality of API documentation.
 
@@ -112,10 +116,7 @@ Elixir documents FFI bindings, see https://hexdocs.pm/rayex/Rayex.Core.html#begi
 It does not have a method for documenting private methods, see https://hexdocs.pm/elixir/1.12/writing-documentation.html#documentation-code-comments.
 > Because private functions cannot be accessed externally, Elixir will warn if a private function has a @doc attribute and will discard its content.
 
-# Unresolved questions
-
-TBD
-
 # Future possibilities
 
-TBD
+- A `semantic` crystal tool that outputs all Crystal objects regardless of directives
+- More documentation directives such as `:section: Section Name` or `:include: path/to/file.md`
--- a/text/0011-extending-api-docs.md
+++ b/text/0011-extending-api-docs.md
@@ -1,7 +1,7 @@
 - Feature Name: extending-api-docs
 - Start Date: 2024-11-14
 - RFC PR: [crystal-lang/rfcs#11](https://github.com/crystal-lang/rfcs/pull/11)
-- Issue: [crsytal-lang/crystal#6721](https://github.com/crystal-lang/crystal/issues/6721)
+- Issue: [crystal-lang/crystal#6721](https://github.com/crystal-lang/crystal/issues/6721)
 
 # Summary
 
--- a/text/0011-extending-api-docs.md
+++ b/text/0011-extending-api-docs.md
@@ -58,8 +58,8 @@ If a namespace has the `:nodoc:` directive, then the `:showdoc:` directive will 
 ```crystal
 # :nodoc:
 struct MyStruct
-  # This will not show up in API docs
   # :showdoc:
+  # This will not show up in API docs
   struct MyStructChild
   end
 end
--- a/text/0011-extending-api-docs.md
+++ b/text/0011-extending-api-docs.md
@@ -13,6 +13,7 @@ Currently, API documentation is not generated for private/protected methods/obje
 This was originally done as these (typically) should not be used, however, this is not always the case.
 When inheriting from a class that has a protected method that is intended to be implemented, it is useful
 to know that method exists, and what parameters / types it has, without needing to refer to the source code.
+
 Another use case is for libraries such as [raylib.cr](https://github.com/sol-vin/raylib-cr), where developing a
 "Crystal" interface to them using classes and structs would be prohibitive, and currently requires diving
 into the source code in order to figure out what methods are available.
--- a/text/0011-extending-api-docs.md
+++ b/text/0011-extending-api-docs.md
@@ -5,7 +5,7 @@
 
 # Summary
 
-We propose the addition of a `:showdoc:` directive that will allow to document normally undocumented types and methods.
+Introduce a `:showdoc:` directive for the doc generator which includes normally undocumented types and methods in the API documentation.
 
 # Motivation
 
--- a/text/0011-extending-api-docs.md
+++ b/text/0011-extending-api-docs.md
@@ -21,7 +21,7 @@ into the source code in order to figure out what methods are available.
 # Guide-level explanation
 
 The `:showdoc:` directive can be added to private or protected objects, as well as C lib binding objects, to have them show up in API documentation.
-By default, these are hidden and should only be shown if they're intended to be used directly.
+By default, these are hidden and should only be shown if they're explicitly intended to be used.
 
 In this example, when generating API documentation, `Foo.foo` will be included even though it is a private method.
 
--- a/text/0011-extending-api-docs.md
+++ b/text/0011-extending-api-docs.md
@@ -5,8 +5,7 @@
 
 # Summary
 
-This RFC would add a directive `:showdoc:` that would allow private / protected methods and objects,
-as well as C lib, fun, enum, etc, to show up in API documentation.
+We propose the addition of a `:showdoc:` directive that will allow to document normally undocumented types and methods, including libs and external functions.
 
 # Motivation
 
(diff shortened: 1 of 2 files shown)


jgaskins:
Strongly in favor of this. I agree with some of the commenters on the forum that documenting methods that nobody but the library authors will ever use is not a great idea, but sometimes the intended audience of a `protected` method isn't you, but it also isn't _everyone_. People need to be able to see what methods are available to them in the context they're developing in.

In my case, I have an ORM called [Interro](https://github.com/jgaskins/interro) where your query objects inherit from a framework type. Methods like `where` and `order_by` are `protected`, but that's not because Interro uses them internally — it doesn't use them at all. Interro implements them so that your query objects can use them, but they're `protected` so that the rest of your application can't be coupled to your DB schema. Here is a quick example:

```crystal
struct PostQuery < Interro::QueryBuilder(Post)
  def by(author : User)
    where author_id: author.id
  end

  def published
    where { |post| post.published_at < Time.utc }
  end

  def in_reverse_chronological_order
    order_by created_at: :desc
  end
end
```

In this example, note that I'm implementing my query object's methods in terms of `where` and `order_by`. The way I use the `PostQuery` is to use these methods:

```crystal
author_feed = PostQuery.new
  .by(author)
  .published
  .in_reverse_chronological_order
```

If I were able to call `where` or `order_by` from outside of `PostQuery` (for example, in the endpoint that renders the feed), my application would become coupled to the specifics of my DB schema. Since they're `protected`, I maintain a layer of separation so that if I change the mechanics of what it means for a post to be `published` or if I add the ability for multiple users to be listed as authors of a post, I can isolate those changes to these methods and the rest of my application does not need to change.

Application developers should be able to see what methods are available to them inside their query objects, so it's important to be _able_ to document them at the very least.

ysbaddaden (commented):
Thank you for this proposal :bow:

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+If a namespace has the `:nodoc:` directive, then the `:showdoc:` directive will have no effect on anything in its namespace.
+
+```crystal
+# :nodoc:
+struct MyStruct
+  # This will not show up in API docs
+  # :showdoc:
+  struct MyStructChild
```
straight-shoota, line 63: **suggestion:** `:nodoc:` must be the first thing at the start of the comment, otherwise it has no effect.
I think the same should apply for `:showdoc:`. We should spell that out in writing and show it in the examples, maybe adding a line of doc comment below every `:showdoc:`?

The order in this example needs to be changed to be meaningful.
```suggestion
  # :showdoc:
  # This will not show up in API docs
  struct MyStructChild
```

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,96 @@
+- Feature Name: extending-api-docs
+- Start Date: 2024-11-14
+- RFC PR: [crystal-lang/rfcs#11](https://github.com/crystal-lang/rfcs/pull/11)
+- Issue: [crystal-lang/crystal#6721](https://github.com/crystal-lang/crystal/issues/6721)
+
+# Summary
+
+We propose the addition of a `:showdoc:` directive that will allow to document normally undocumented types and methods, including libs and external functions.
```
straight-shoota, line 8: **polish:** The latter part feels unnecessary for the summary. I don't mind either way though.

```suggestion
We propose the addition of a `:showdoc:` directive that will allow to document normally undocumented types and methods.
```

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,96 @@
...
+
+# Motivation
+
+Currently, API documentation is not generated for private/protected methods/objects or C lib binding objects.
+This was originally done as these (typically) should not be used, however, this is not always the case.
+When inheriting from a class that has a protected method that is intended to be implemented, it is useful
+to know that method exists, and what parameters / types it has, without needing to refer to the source code.
+
+Another use case is for libraries such as [raylib.cr](https://github.com/sol-vin/raylib-cr), where developing a
+"Crystal" interface to them using classes and structs would be prohibitive, and currently requires diving
```
straight-shoota, line 18: **polish:** Clarify the relevance with respect to undocumented objects.

```suggestion
"Crystal" interface to them using classes and structs would be prohibitive, and expose the lib bindings as part of the public API. It currently requires diving
```

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,96 @@
...
+into the source code in order to figure out what methods are available.
+
+# Guide-level explanation
+
+The `:showdoc:` directive can be added to private or protected objects, as well as C lib binding objects, to have them show up in API documentation.
+By default, these are hidden and should only be shown if they're intended to be used directly.
```
straight-shoota, line 24: **polish:** IMO _explicit_ intention to be used is more appropriate here. Or maybe I'm misinterpeting the sentence? 🤔 

```suggestion
By default, these are hidden and should only be shown if they're explicitly intended to be used.
```

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,96 @@
...
+end
+```
+
+# Reference-level explanation
+
+- The parser will need to be updated to support doc comments for C lib binding objects and the `:showdoc:` directive
+- The documentation generator will need to be updated to support C lib binding objects and private/protected objects
+- If an object has a `:showdoc:` directive and its parent namespace is shown, then it should be shown too
+
+TBD
```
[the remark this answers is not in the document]
straight-shoota: This seems already pretty exhaustive to me, because the subject is relatively straightforward to explain and doesn't have too much depth to it (in part because it's close analogy with the established `:nodoc:` semantics).

Technical issues are only relevant here if they affect the proposed behaviour in a way that makes parts of it unable to work.
I don't see much risks for hitting any such obstables, if there are any technical challenges the'll probably just be implementation details.

text/0011-extending-api-docs.md
nobodywasishere: Ahh okay that's true of the existing `:nodoc:` as well. I personally prefer it being the last line but that's fine with me. We'll also need to strip this in the docs generator so it isn't printed out (if that doesn't happen already with directives).

straight-shoota: Yeah it'll need to be stripped explicitly. Currently there is no general concept of directives, the doc generator just skips if the doc string starts with `:nodoc:` and no further processing is needed.

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,96 @@
...
+# Rationale and alternatives
+
+The other design that has been considered is having flags on the documentation generator itself that enable showing of private / protected objects and C lib objects in the API documentation. We chose not to go with this design as it required flags to be added at generation time, and only generated all or none (no granularity in what is shown).
+
+This cannot be done in a library instead as it requires updates to the parser itself. This proposal makes Crystal code easier to understand, as it increases the amount and quality of API documentation.
+
+# Prior art
+
+There is a [PR](https://github.com/crystal-lang/crystal/pull/14816) implementing a similar feature, however it uses the generation-time flag method mentioned above, instead of the `:showdoc:` directive.
+
```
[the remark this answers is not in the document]
nobodywasishere: I've added some quick notes on what Ruby, Rust, and Elixir do for FFI bindings and private methods.

text/0011-extending-api-docs.md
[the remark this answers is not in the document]
ysbaddaden: As I said: not much. The details explained above are already a lot :)

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,121 @@
...
+# Motivation
+
+Currently, API documentation is not generated for private/protected methods/objects or C lib binding objects.
+This was originally done as these (typically) should not be used, however, this is not always the case.
+When inheriting from a class that has a protected method that is intended to be implemented, it is useful
+to know that method exists, and what parameters / types it has, without needing to refer to the source code.
+
+Another use case is for libraries such as [raylib.cr](https://github.com/sol-vin/raylib-cr), where developing a
+"Crystal" interface to them using classes and structs would be prohibitive, and expose the lib bindings as part of the public API. It currently requires diving
```
RX14, line 18:
```suggestion
"Crystal" interface to them using classes and structs would be prohibitive, therefore they expose the lib bindings as part of the public API. It currently requires diving
```

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,121 @@
...
+- Issue: [crystal-lang/crystal#6721](https://github.com/crystal-lang/crystal/issues/6721)
+
+# Summary
+
+We propose the addition of a `:showdoc:` directive that will allow to document normally undocumented types and methods.
+
+# Motivation
+
+Currently, API documentation is not generated for private/protected methods/objects or C lib binding objects.
+This was originally done as these (typically) should not be used, however, this is not always the case.
```
RX14, line 13:
```suggestion
This was originally done as these (typically) should not be used externally, however, this is not always the case.
```

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,121 @@
...
+When inheriting from a class that has a protected method that is intended to be implemented, it is useful
+to know that method exists, and what parameters / types it has, without needing to refer to the source code.
+
+Another use case is for libraries such as [raylib.cr](https://github.com/sol-vin/raylib-cr), where developing a
+"Crystal" interface to them using classes and structs would be prohibitive, and expose the lib bindings as part of the public API. It currently requires diving
+into the source code in order to figure out what methods are available.
+
+# Guide-level explanation
+
+The `:showdoc:` directive can be added to private or protected objects, as well as C lib binding objects, to have them show up in API documentation.
```
RX14, line 23:
```suggestion
The `:showdoc:` directive can be added to private or protected objects, methods, and C lib bindings, to have them show up in API documentation.
```

RX14:
Definitely in favour of this proposal, it seems useful, and the design is in keeping with the rest of the doc generator. One thing to keep an eye on is that the doc generator makes the visibility (protected/private) of the method clear and obvious in the generated docs. It should be impossible to miss.

sol-vin:
If I can suggest one possibility, when a `lib` `fun` gets documented can it include the name of the c function it's binding in the docs as well.

nobodywasishere:
> Definitely in favour of this proposal, it seems useful, and the design is in keeping with the rest of the doc generator. One thing to keep an eye on is that the doc generator makes the visibility (protected/private) of the method clear and obvious in the generated docs. It should be impossible to miss.

Other people in the past have mentioned having separate sections for private and protected methods, would that work? My previous implementation just put the visibility next to the method name (as it would look in code)

RX14:
> Other people in the past have mentioned having separate sections for private and protected methods, would that work? My previous implementation just put the visibility next to the method name (as it would look in code)

I'm ambivalent about having a separate section, I mostly wanted to make a point about the visual distinctiveness of the visibility modifier when looking at a single method. You can't rely on sections for this because the section is context you can forget if the section marker is off-screen, and therefore doesn't contribute to being immediately obvious.

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,121 @@
...
+- The documentation generator will need to be updated to support C lib binding objects and private/protected objects
+- If an object has a `:showdoc:` directive and its parent namespace is shown, then it should be shown too
+
+TBD
+
+# Drawbacks
+
+TBD
+
+# Rationale and alternatives
```
straight-shoota, line 86: Another consideration is a more generic syntax for doc metadata. This could at least to be mentioned as alternative.

There are ideas for other enhancements that could be attached to doc comments, such as section/group information (https://github.com/crystal-lang/crystal/issues/1312). And I think there are many more use cases, though mostly unexplored yet.

At some point it would certainly be necessary to have a more sophisticated syntax for such data markup. An insipiration could be Crystal's annotation syntax, for example.

I suppose venturing out on this would blow up the scope and put an unnecessary delay on this RFC. It's useful in itself and can be implemented quickly, without much disruption. And it builds on already existing syntax in form of `:nodoc:`.

nobodywasishere: > An insipiration could be Crystal's annotation syntax

I mean `:nodoc:` and `:showdoc:` could've been built using annotations instead of comment directives, though that wouldn't work for directives like `:section:`.

Honestly the current syntax is reminiscent of rST's [field lists](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#field-lists), could be a good source of inspiration for what we could do here.

nobodywasishere:
Is there anything else to do for this RFC to be merged? Want to get started on this for Crystal 1.15 if possible.

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,122 @@
+- Feature Name: extending-api-docs
+- Start Date: 2024-11-14
+- RFC PR: [crystal-lang/rfcs#11](https://github.com/crystal-lang/rfcs/pull/11)
+- Issue: [crystal-lang/crystal#6721](https://github.com/crystal-lang/crystal/issues/6721)
+
+# Summary
+
+We propose the addition of a `:showdoc:` directive that will allow to document normally undocumented types and methods.
```
straight-shoota, line 8: **polish:** This could perhaps be a bit more expressive.
```suggestion
Introduce a `:showdoc:` directive for the doc generator which includes normally undocumented types and methods in the API documentation.
```

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,122 @@
...
+
+# Motivation
+
+Currently, API documentation is not generated for private/protected methods/objects or C lib binding objects.
```
straight-shoota, line 12: **polish:** Perhaps it would be helpful to use a single term for documented things (methods, types, constants, macros, functions, ...). I think *"object"* fits great. Maybe we could introduce that here and mention what it refers to, and then later occurences in the text could simplify to the term "(documented) objects"? (This is already happening in some places, actually)

For example:
```diff
@@ -23,1 +23,1 @@
-The `:showdoc:` directive can be added to private or protected objects, methods, and C lib bindings, to have them show up in API documentation.
+The `:showdoc:` directive can be added to `private` or `protected` objects, and objects in `lib` bindings, to have them show up in API documentation.
```

text/0011-extending-api-docs.md
```diff
@@ -0,0 +1,122 @@
...
+This was originally done as these (typically) should not be used externally, however, this is not always the case.
```
straight-shoota, line 13: **polish:** 
```suggestion
This was originally done as these (typically) should not be used externally. However, this is not always the case.
```

straight-shoota (approved):
LGTM.
I only have some minor suggestions to polish the language. But none of them are blockers.

The general is well received, so I think we can already start on the implementation and finish the RFC in parallel.

text/0011-extending-api-docs.md
ysbaddaden: I'm not sure about the "object" word. It's very connoted for me as type instances and has zero relationship with methods or C symbols :confused:

straight-shoota: Maybe "features" then?

straight-shoota (approved):
I think we can merge this. Everything has been implemented in https://github.com/crystal-lang/crystal/pull/15337 and https://github.com/crystal-lang/crystal/pull/15447

text/0011-extending-api-docs.md
mverzilli: elements? constructs?

PR #12, opened 2024-11-22
RFC 0012: timers

ysbaddaden:


text/0012-timers.md
```diff
@@ -0,0 +1,310 @@
...
+  # Returns the absolute time at which the next expiring timer is scheduled at.
+  # Returns nil if there are no timers.
+  abstract def next_expiring_at? : Time::Span?
+end
+```
+
+## Data structure: min pairing heap
+
+A min-heap is a simple, fast and efficient tree data structure, that keeps the
+smaller value as the HEAD of the tree (the rest isn't ordered). This is enough
```
RX14, line 148: smallest?

text/0012-timers.md
```diff
@@ -0,0 +1,310 @@
...
+
+# Prior art
+
+- `libevent` stores events with a timer into a min-heap, but it also keeps a
+  list of "common timeouts"... I didn't investigate what they mean by it
+  exactly.
+
+- Go stores all timers into a min-heap (4-ary) but allocates timers in the GC
+  HEAP and merely marks cancelled timers on delete. I didn't investigate how
+  it deals with the tombstones.
```
RX14, line 285: This sounds like the best design to me, I actually wrote a message above which described doing this, until I read this part. Tombstones can probably be kept around until dequeue, though there may be other opportune times to delete them if scanning/moving entries anyway.

ysbaddaden: It could be interesting to understand, but I wonder about the benefit.

Keeping the tombstones means they might stay for seconds, minutes or hours despite having been cancelled. The occupancy would no longer be how many active timers there are now, but the total number of timers created in the last N seconds/minutes/hours.

They also increase the cost of delete-min: it must be repeated multiple times until we reach a not cancelled timer (not cool).

We'd have to allocate the event in the GC HEAP (we currently allocate events on the stack) and they'd stay allocated until they finally leave the 4-heap.

We can probably clear the tombstones we meet as we swap items (cool), but that means dereferencing each pointer, which reduces the CPU cache benefit of the flat array...

RX14: Yes, the practicalities of the solution might outweigh any performance benefit, but Go leaning that way provides signal to me it's worth doing the benchmarking.

text/0012-timers.md
```diff
@@ -0,0 +1,310 @@
...
+nodes) led to to slighly faster `insert` time than a D-ary Heap (that needs more
+swaps) especially when timers come out of order, but a noticeably slower
+`delete-min` since it must rebalance the tree. The `delete` operation however
+quickly outperforms the 4-heap, even at low occupancy (a hundred timers) and
+never balloons.
+
+Despite the drawback on the `delete-min` operation, a benchmark using mixed
+operations (insert + delete-min, insert + delete) led the pairing heap to have
+the best overall performance. See the [benchmark
+results](https://gist.github.com/ysbaddaden/a5d98c88105ea58ba85f4db1ed814d70à)
```
vlazar, line 167: Extraneous à at the end of the link
```suggestion
results](https://gist.github.com/ysbaddaden/a5d98c88105ea58ba85f4db1ed814d70)
```

text/0012-timers.md
```diff
@@ -0,0 +1,310 @@
...
+we need to handle the correct execution of timers ourselves.
+
+We must handle timers, we must store them into efficient data structure(s), and
+we must suppor the following operations:
+
+- create a timer;
+- cancel a timer;
+- execute expired timers;
+- determine the next timer to expire, so we can decide for how long a process or
+  thread can be suspended (usually when there is nothing to do).
```
yxhuvud, line 24: The need for this depends on the event loop. Is there any need at all for all this complexity if the event loop is driven by io_uring? Just emit a timeout event for each fiber that is waiting, and that's it. I'm all for shared code between underlying event loops that are limited in what they can do, but are there any reason to lock event loops into more structure?

It also supports timeouting io operations.

ysbaddaden: `io_uring` supporting timeouts on IO operations => :heart: 

Each eventloop can use whatever it pleases, yet... there are still "select action timeouts" that can be cancelled, so even `io_uring` will need to support an arbitrary dequeue of timeouts.

Even with events to notify the blocking waits (which we use for epoll, kqueue and IOCP), we still need to rearm the timer after it triggered (for example) and need to know when the next timer is expiring. I don't think `io_uring` will be treated differently.

So far, my naive vision is for `io_uring` to notify an `eventfd` registered to an `epoll` instance, along with a `timerfd` (for precise timers) and eventually more niceties (i.e. `signalfd` and `pidfd`).

yxhuvud: > Each eventloop can use whatever it pleases, yet... there are still "select action timeouts" that can be cancelled, 

Yes. For example using the uring op TIMEOUT_REMOVE. 

> so even io_uring will need to support an arbitrary dequeue of timeouts.

No, that does not follow. It may be an issue if we are not ok waiting for the response to the timeout removal I guess, and it also need to handle the race condition where the timer is already triggering and execute before the actual timeout removal.

> `timerfd`  (for precise timers),

FWIW, uring timeout op also take timespec structs as arguments with the same precision as timerfd. What uring doesn't seem to support is the periodic part of the argument, but instead there is an MULTISHOT argument if you want repeating triggers.

> So far, my naive vision is for io_uring to notify an eventfd registered to an epoll instance

I'd suggest not using epoll at all and instead use the uring POLL op, which does more or less the same but a lot simpler. 


But in any case I guess it doesn't matter too much as it doesn't really impact the public interfaces so in the end it can be changed when the need arises..

ysbaddaden: Thanks for all the information! I'll have to dig much deeper into `io_uring` capabilities.

AFAIK all timeouts in the Linux kernel go into the timer wheel (tick based, low precision, now loses precision over time) while timers go into hrtimer (high precision, no ticks, nanoseconds clock).

I'd expect io_uring timeouts to end up into the timer wheel, which is fine for timeouts, but I'd like to keep `timerfd` for `sleep(seconds)`.

https://www.kernel.org/doc/html/latest/timers/highres.html

text/0012-timers.md
```diff
@@ -0,0 +1,310 @@
...
+timers in the Crystal runtime.
+
+# Motivation
+
+With the Event Loop overhaul made possible by [RFC 7] and achieved in [RFC 9]
+where we remove the libevent dependency, that we already didn't use on Windows,
+we need to handle the correct execution of timers ourselves.
+
+We must handle timers, we must store them into efficient data structure(s), and
+we must suppor the following operations:
```
vlazar, line 18:
```suggestion
we must support the following operations:
```

text/0012-timers.md
ysbaddaden: It could be a cleanup during a GC collection :thinking:

text/0012-timers.md
```diff
@@ -0,0 +1,310 @@
...
+  # Returns the absolute time at which the next expiring timer is scheduled at.
+  # Returns nil if there are no timers.
+  abstract def next_expiring_at? : Time::Span?
+end
+```
+
+## Data structure: min pairing heap
+
+A min-heap is a simple, fast and efficient tree data structure, that keeps the
+smaller value as the HEAD of the tree (the rest isn't ordered). This is enough
```
ysbaddaden, line 148:
```suggestion
smallest value as the HEAD of the tree (the rest isn't ordered). This is enough
```

text/0012-timers.md
```diff
@@ -0,0 +1,310 @@
...
+   The low resolution is because timeouts are mostly about bookkeeping, to
+   eventually close a connection after some time has passed for example, so a
+   10s timeout running after 11s won't create issues.
+
+2. **Timers**: Timers are used to schedule ongoing events. They can have high
+   resolution requirements, and usually expire.
+
+   In crystal such a `timer` is created when we call `sleep(Time::Span)` or
+   `Fiber.yield` that behaves as a `sleep(0.seconds)`. There are no public API
+   to cancel a sleep, and they always expire.
```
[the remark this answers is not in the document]
ysbaddaden:
```suggestion
   In crystal such a `timer` is created when we call `sleep(Time::Span)`.
   There is no public API to cancel a sleep, and they always expire.
```

text/0012-timers.md
```diff
@@ -0,0 +1,310 @@
...
+
+   - low precision (some milliseconds is acceptable);
+   - fast `delete` operation (likely to be cancelled);
+   - must accomodate many timeouts at any given time (e.g. c10k problem).
+
+2. Timers (sleeps):
+
+   - high precision (sub-millisecond and below is desireable);
+   - no need for `delete` (never cancelled);
+   - more reasonable number of timers (**BOLD CLAIM TO BE VERIFIED**)
```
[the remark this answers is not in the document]
ysbaddaden:
```suggestion
   - more reasonable number of timers compared to timeouts (**BOLD CLAIM TO BE VERIFIED**)
```

PR #14, opened 2025-08-01
RFC XXXX: General Timeouts

ysbaddaden:
Synchronization primitives, such as mutexes, condition variables, pools, or event channels, could take advantage of a general timeout mechanism.

Crystal has a mechanism in the event loop to suspend the execution of a fiber for a set amount of time (`#sleep`). It also has a couple mechanisms to add timeouts: one associated to IO operations, and another tailored to `Channel` and `select` to support the timeout branch of select actions.

Adding timeouts to all the synchronization primitives in the stdlib, and possibly to custom ones in shards and applications, shouldn't be much harder than calling `#sleep`, or need to hack into the private `Fiber#timeout_event`.

### DIFF
--- a/text/0000-general-timeout.md
+++ b/text/0000-general-timeout.md
@@ -1,0 +1,226 @@
+---
+Feature Name: general-timeout
+Start Date: 2025-08-02
+RFC PR: "https://github.com/crystal-lang/rfcs/pull/0000"
+Issue: N/A
+---
+
+# Summary
+
+Introduce a general `#timeout` feature, ideally as simple as `#sleep` that would
+return whether the timer expired, or was manually canceled.
+
+
+# Motivation
+
+Synchronization primitives, such as mutexes, condition variables, pools, or
+event channels, could take advantage of a general timeout mechanism.
+
+Crystal has a mechanism in the event loop to suspend the execution of a fiber
+for a set amount of time (`#sleep`). It also has a couple mechanisms to add
+timeouts: one associated to IO operations, and another tailored to `Channel` and
+`select` to support the timeout branch of select actions.
+
+> [!CAUTION]
+> Verify if the select action timeout mechanism can requme *twice*.
+
+Adding timeouts to all the synchronization primitives in the stdlib, and
+possibly to custom ones in shards and applications, shouldn't be much harder
+than calling `#sleep`, or need to hack into the private `Fiber#timeout_event`.
+
+
+# Guide-level explanation
+
+The complexity of timeout over sleep is that it can be canceled, which leads to
+synchronization issues. For example, multiple threads may try to cancel the
+timeout at the same time, and yet another thread might try to process the
+expired timer too.
+
+We want the fiber to be resumed *once* and to report whether the timeout expired
+or was canceled. We need synchronization and a mean to decide which thread
+resolves the timeout and will enqueue the fiber. There can of course be only
+one.
+
+A straightforward solution is to use an atomic: the fiber sets the atomic before
+suspending itself, then any attempt to enqueue the fiber must be the one that
+succeeds to unset the atomic. This can be achieved with an `Atomic(Bool)` for
+example (`#set(true)` then `#swap(false) == true`).
+
+The atomic could be a property on `Fiber`, it's easily accessible to anything
+that already knows about it, be it a waiting fiber in a mutex or a waiting timer
+event in the event loop. The problem is that there is no telling if the timeout
+changed in between. This is more commonly known as the ABA problem.
+
+The common solution is to increment the atomic value on every attempt to change
+the value using compare-and-set (CAS). If the atomic value didn't change, the
+CAS succeeds, and the timeout is succesfully resolved. Other attempts will fail
+because the value changed, so even if the timeout is set again, the value will
+have been incremented and the CAS will always fail. Albeit, the counter will
+eventually wrap, but after so many iterations that it's impossible to hit the
+ABA issue in practice. For example, an UInt32 word with a 1 bit flag would need
+2,147,483,648 iterations!
+
+The counterpart is that every waiter and timer event must know the actual atomic
+value, thereafter called `cancelation token`, to be able to resolve the timeout.
+
+> [!NOTE]
+> Alternatively, we could allocate the atomic in the HEAP, but then every
+> timeout would depend on the GC, and we'd still need to save the pointer for
+> every waiter and timer. We can't store it on `Fiber` because we'd jump back
+> into the ABA problem.
+
+## Synchronization primitives
+
+Let's take a mutex for the example.
+
+On failure to acquire the lock, the current fiber will want to add itself into a
+waiting list (private to the mutex). As explained in the previous section, we
+must memorize the cancelation token to be able to cancel it, so it must set the
+timeout, get the token, and then add the fiber and the token to the mutex
+waiting list, then delegate the timeout to the event loop implementation.
+
+When another fiber unlocks, the mutex must try to wake a waiting fiber. While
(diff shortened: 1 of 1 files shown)


text/0000-general-timeout.md
```diff
@@ -0,0 +1,226 @@
...
+    # token to a waiting list before the fiber suspends itself
+    yield token
+
+    expired = Crystal::EventLoop.current.timeout?(duration, token)
+    expired ? TimeoutResult::EXPIRED : Timeout::CANCELED
+  end
+
+  # :nodoc:
+  protected def set_timeout : TimeoutToken
+    token = (@timeout.get(:relaxed) &+ 2_u32) | 1_u32
```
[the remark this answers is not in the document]
ysbaddaden: The `|` is adding the flag (bit 1) while `&+` is incrementing the counter (bits 2..). I should explain that.

ysbaddaden:
@beta-ziliani Ah, I pushed a new draft between your review. I removed the implementation details to focus on the API, and added a technical example to the reference section for the mutex from the guide section.

text/0014-general-timeouts.md
```diff
@@ -0,0 +1,313 @@
...
+Feature Name: general-timeouts
+Start Date: 2025-08-02
+RFC PR: "https://github.com/crystal-lang/rfcs/pull/0014"
+Issue: N/A
+---
+
+# Summary
+
+Introduce a general `#timeout` feature, ideally as simple as `#sleep` that would
+return whether the timer expired, or was manually canceled.
```
straight-shoota, line 11: **suggestion:** The term `timeout` is not very clearly defined. We're already using it for a number of slightly different things. Hence we should clarify what "`#timeout` feature" actually means. Otherwise it may take halfway through the guide-level explanation to finally grasp it.

I understand the concept is like sleep but with a handle for cancelling it, which may be triggered from a different fiber in order to interrupt the sleeping fiber.

text/0014-general-timeouts.md
```diff
@@ -0,0 +1,313 @@
...
+
+> [!CAUTION]
+> Verify if the select action timeout mechanism can resume *twice*.
+
+Adding timeouts to all the synchronization primitives in the stdlib, and
+possibly to custom ones in shards and applications, shouldn't be much harder
+than calling `#sleep`, or need to hack into the private `Fiber#timeout_event`.
+
+
+# Guide-level explanation
```
straight-shoota, line 32: **suggestion:** This section feels a bit blown up and missing the point. The idea of the guide-level explanation is to give an high-level overview of the final state of the proposal. It should teach Crystal developers how to use this feature. Not how we decided the implementation details. Discussions about the approach and alternatives should go into the *Rationale* section, for example.
See https://github.com/crystal-lang/rfcs/blob/main/0000-template.md for hints

There should be some example code for illustration.

text/0014-general-timeouts.md
```diff
@@ -0,0 +1,313 @@
...
+
+2. We may want to use *absolute time instead of relative duration*, so every use
+   cases that need to retry wouldn't have to deal with caculating the remaining
+   time on each iteration, and would only need to calculate the absolute limit
+   (now + timeout).
+
+   The problem is that monotonic times and durations are represented using the
+   same type (`Time::Span`), and `#sleep` uses relative time. Using absolute
+   time would be confusing. Maybe we can support *both* and add an `absolute`
+   argument, that would default to `false`?
```
straight-shoota, line 292: **thought:** Instead of a separate flag parameter, we could use separate defs with different named parameters.

```cr
def Fiber.timeout(duration : Time::Span, & : Fiber::TimeoutToken) : Fiber::TimeoutResult
end

def Fiber.timeout(*, until : Time::Span, & : Fiber::TimeoutToken) : Fiber::TimeoutResult
end
```

ysbaddaden:
@straight-shoota I renamed the RFC to "cancelable timers", clarified the summary/motivation sections, and reorganized the sections. Is it better?

I made the rationale a distinct section, and put it between the guide and reference sections, so we flow from motivation -> explanation with example -> how to implement the timeout -> proposed API -> code example for the explanation (based on how + api).

text/0014-cancelable-timers.md
```diff
@@ -0,0 +1,342 @@
...
+[sync shard].
+
+We might be able to use it in the IOCP event loop where we currently need to
+sleep and yield in a specific timeout case, as well as other event loops for IO
+timeouts (to be verified), as well as for the select action timeout that
+currently relies on a custom solution.
+
+[RFC 0007]: https://github.com/crystal-lang/rfcs/pull/7
+[RFC 0012]: https://github.com/crystal-lang/rfcs/pull/12
+[`sleep(time)`](https://crystal-lang.org/api/1.17.1/toplevel.html#sleep(time:Time::Span):Nil-class-method)
```
ysbaddaden, line 339:
```suggestion
[`sleep(time)`]: https://crystal-lang.org/api/1.17.1/toplevel.html#sleep(time:Time::Span):Nil-class-method
```

ysbaddaden:
I included your suggestions.

text/0014-cancelable-timers.md
```diff
@@ -0,0 +1,356 @@
...
+
+The public API can do with a couple methods:
+
+- `Fiber.timeout(duration : Time::Span, & : Fiber::TimeoutToken ->) : Fiber::TimeoutResult`
+
+  Sets the flag and increments the value of the atomic. Yields the cancelation
+  token (aka the new atomic value) so the caller can record it (the block is a
+  called-before-suspend callback), then delegates to the event loop to suspend
+  the calling fiber and eventually resume it when the timeout expires, if it
+  hasn't been canceled already.
```
straight-shoota, line 175: Looking at [the spec for cancel behaviour](https://github.com/ysbaddaden/crystal/blob/a9919abc57758af170f311e9aff88b2110f605bc/spec/std/fiber_spec.cr#L36-L54), I'm wondering about a synchronization issue: What happens when the token is picked up and canceled _before_ the event loop has actually set up the timer and suspended the fiber?

The token is intended to be passed to a different fiber. Some fiber running in a different thread might just be waiting for it and immediately resolve it. The timer could not possibly have expired, so it should return true. But then it would try resuming the fiber while it's till running...?

text/0014-cancelable-timers.md
```diff
@@ -0,0 +1,356 @@
...
+
+# Prior art
+
+Most runtimes expose cancelable timers, directly (for example [timer_create(2)]
+on POSIX) or indirectly through synchronization primitives (for example
+[pthread_cond_timedwait(3)] on POSIX).
+
+# Unresolved questions
+
+1. Instead of adding a distinct `Fiber.timeout(time)` in addition to
```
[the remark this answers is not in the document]
ysbaddaden: Yes. Both `sleep` and `timeout` could just be cancelable timers internally.

ysbaddaden:
Returning to this with a fresh :brain: and following @beta-ziliani's remarks, I think we could have:

- `Fiber::CancelationToken` (instead of `TimeoutToken`) so we can internally reuse the mechanism for `select` regardless of the presence of a timeout action;

- `Fiber.sleep(Time::Span, & : CancelationToken -> TimeoutResult)` instead of `.timeout`;

- `::sleep(Time::Span)` becomes a shortcut for `Fiber.suspend(Time::Span) { }`

- add `Crystal::EventLoop#sleep(Time::Span, CancelationToken)` to replace the existing method instead of introducing a new `#timeout` method.

I'm leaving aside the absolute timers because they're relative to a specific clock and need to consider that first.

ysbaddaden:
Both the RFC and the implementation PR have been updated to reflect the above proposal.

PR #15, opened 2025-08-28
RFC: `Time::Monotonic`

straight-shoota:


text/0000-time-monotonic.md
```diff
@@ -0,0 +1,175 @@
...
+- The name for the type is not settled. Alternatives:
+  - `Time::Monotonic`: Concise and explicit about clock domain, but less precise
+    about _what_ it is. Aligns with existing `Time.monotonic`.
+  - `Time::Instant`: More generic, leaves room for multiple backends.
+  - ?
+- Monotonic clocks can have different characteristics about whether they
+  progress while the system is suspended. Should there be support for different
+  clocks? See [`CLOCK_MONOTONIC` vs
+  `CLOCK_BOOTTIME`](https://man7.org/linux/man-pages/man2/timerfd_create.2.html).
+  This affects the naming decision.
```
ysbaddaden, line 127: Yes, `CLOCK_BOOTTIME` can indeed be very interesting since it counts suspended time.

There's also `CLOCK_UPTIME` on BSD, and the [OpenBSD manpage](https://man.openbsd.org/clock_gettime.2) has a nice example: you can do boottime - uptime to determine for how long the system has been suspended since the initial boot, meaning that cross clock substractions of instants can sometimes be valid.

HertzDevil: For the record, Windows has only the equivalent of `CLOCK_BOOTTIME`.

straight-shoota: Oh right. I suppose that means `Time.monotonic` already behaves differently on Windows and Unix systems? 🙈 
`QueryPerformanceCounter` counts while the system is suspended, `CLOCK_MONOTONIC` does not.

Should we switch to `CLOCK_BOOTTIME` then for the default clock on Unix? 🤔

text/0000-time-monotonic.md
```diff
@@ -0,0 +1,175 @@
...
+  [#3827])
+- Extract commonalities between `Time` and `Time::Monotonic` into a module type
+  (`Time::Instant`?).
+- Unified clock APIs: e.g., `sleep(until: Time::Monotonic)` and `sleep(until:
+  Time)` (see _[Generalize `#sleep` for monotonic and wall clock]_)
+- Constructor from a raw value, and a converter to a raw value. The use case
+  would be to import/export clock readings. This is primarily useful for
+  interacting with components outside stdlib (C libraries). Serialization
+  probably isn't much relevant though, because the values are really only valid
+  inside the current process.
```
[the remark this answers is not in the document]
ysbaddaden: Well, we need to export the data so we can use it and be interoperable. For example pass it to an arbitrary POSIX function that takes a timespec or to some Windows library through some other struct.

text/0000-time-monotonic.md
```diff
@@ -0,0 +1,175 @@
...
+  This affects the naming decision.
+- The `#-` method for calculating a duration between two instants is simple, but
+  can lead to mistakes based on the order of operands. It's also susceptiple to
+  errors in monotonicity which can lead to negative durations.
+  An explicit distance calculation that ensures monotonicity would be more safe.
+  An example for that is Rust's [`duration_since`]: It clearly specifies the
+  expected temporal relation and saturates the return value at zero.
+  Such a method could be generally useful, but questions the existence of `#-`.
+  `#elapsed` would be convenient alternative that also alleviates these issues,
+  but doesn't cover all use cases.
```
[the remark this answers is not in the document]
straight-shoota: Rust's `Instant.sub` delegates to `duration_since`, so it's also saturated. Maybe we could consider that as well for `#-`?

text/0000-time-monotonic.md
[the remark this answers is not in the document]
ysbaddaden: Then Rust's `std::time::Instant` and Zig's `std.time.Instant` are only useful for measuring elapsed time, and nothing else :confused:

text/0000-time-monotonic.md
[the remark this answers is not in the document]
straight-shoota: No, it's also used for deadlines in [`thread::sleep_until`](https://doc.rust-lang.org/std/thread/fn.sleep_until.html), for example. You can calculate `now + duration` to get an instant in the future. But the API doesn't disclose the system-specific representation of that value.

text/0000-time-monotonic.md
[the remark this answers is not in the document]
straight-shoota: It's also used for timeouts, for example in [`thread::sleep_until`](https://doc.rust-lang.org/std/thread/fn.sleep_until.html). You can calculate `now + duration` to get an instant in the future. But the API doesn't disclose the system-specific representation of that value.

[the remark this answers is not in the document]
ysbaddaden: I guess that works for stdlib usages, but you're left to do it yourself if you need to interact with some other library (even just some libc), or use dirty hacks:

```crystal
struct Time::Monotonic
  ZERO = uninitialized self
  pointerof(ZERO).as(UInt8*).clear(sizeof(self))

  def to_span : Time::Span
    self - ZERO
  end
end
```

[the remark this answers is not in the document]
straight-shoota: My preference would be to leave this out of the initial implementation. We should look more closely into specific use cases. It's easy to enhance the type later, but it would be hard to trim it down.

text/0000-time-monotonic.md
ysbaddaden: Interesting. I agree there's no point to try to get more than REALTIME and MONOTONIC, though things maybe change in 13 years. Here are the manpages for `clock_gettime`:

- [Darwin](https://keith.github.io/xcode-man-pages/clock_gettime.3.html)
- [DragonflyBSD](https://man.bsd.lv/DragonFly-6.0.1/clock_gettime.2)
- [FreeBSD](https://man.freebsd.org/cgi/man.cgi?query=clock_gettime)
- [Linux](https://www.man7.org/linux/man-pages/man2/clock_gettime.2.html)
- [NetBSD](https://man.netbsd.org/clock_gettime.2)
- [OpenBSD](https://man.openbsd.org/clock_gettime.2)

If I'm not mistaken, the following should "monotonically increasing while the system is running (doesn't increase during suspend)":

- Linux: `CLOCK_MONOTONIC` (maybe `_RAW`?)
- Darwin: `mach_absolute_time` (maybe upgrade to `CLOCK_UPTIME`?)
- UNIX: `CLOCK_UPTIME` (with fallback to `CLOCK_MONOTONIC`)
- Win32: `QueryUnbiasedInterruptTime`

Caveat: the manpages state nothing about sleep vs suspend, which might be hardware dependent.

text/0000-time-monotonic.md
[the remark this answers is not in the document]
ysbaddaden: Yeah, let's.

text/0000-time-monotonic.md
straight-shoota: But I think we should use a clock that includes time when the system is suspended. That's means the real elapsed time. Excluding suspended time seems very arbitrary.

straight-shoota: Darwin, Dragonfly and FreeBSD have a clock that counts suspended time. Only NetBSD seems to be missing that.
Hence I mentioned that we cannot make a guarantee that suspended time is included.
But whenever such a clock is available, we should use it. 

Measuring only the time that the system is running is of little practical use. We want to know how much time actually elapsed.

ysbaddaden: Sometimes we want to know how much time elapsed in real life (e.g. timeouts), sometimes with we want to know how much running time has elapsed (e.g. benchmarks).

Both are perfectly valid.

> AFAIK Darwin, Dragonfly and FreeBSD have a clock that counts suspended time.

Links? Because their `clock_gettime` explicitly don't. Are the clocks monotonic (strictly increasing)? are they unaffected by realtime clock changes?

> whenever such a clock is available, we should use it.

Then the behavior isn't portable, and if it's not, just using the monotonic clock as "Strictly Increasing" is enough.

straight-shoota: The manpages of Dragonfly and FreeBSD mention for `CLOCK_UPTIME` that it excludes suspended time.

ysbaddaden: Then the only guarantee is that it's SI. It might not be implemented that way, and even if verified on one version, it might be different in another kernel :shrug:

text/0000-time-monotonic.md
```diff
@@ -0,0 +1,179 @@
...
+transitioning to `Time::Monotonic.now`.
+
+A simple mechanical replacement `s/Time.monotonic/Time::Monotonic.now/` could be
+implemented in tooling (e.g. [ameba](https://github.com/crystal-ameba/ameba)).
+This should generally work fine because the usable API is identical. If the type
+`Time::Span` is encoded in a type restriction (e.g. ivar or method signature),
+it needs updating as well. This is not trivial to automate.
+
+# Reference-level explanation
+
```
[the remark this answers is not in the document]
straight-shoota: See https://mathworld.wolfram.com/StrictlyIncreasingFunction.html and any other definition I could find. "can't go backwards" is weakly increasing.
A constant function is monotonic, but not strictly increasing.

text/0015-time-monotonic.md
```diff
@@ -0,0 +1,220 @@
...
+  but doesn't cover all use cases.
+
+# Future possibilities
+
+- `#elapsed` would allow extensions for different clocks. Calling `#elapsed`
+  would always use the same clock as `self`.
+- A stopwatch / timer implementation, similar to Zig's [`std.time.Timer`] (see
+  [#3827])
+- Extract commonalities between `Time` and `Time::Instant` into a module type
+  (`Time::CockReading`?).
```
Sija, line 192: Hmm... 🐓?

straight-shoota: Well, the idea was using a cock's crow as a metaphor for a natural time source.

Sija: I see, I can just admire how you manage to weave the intricate accents of a rural life into a technical discussion ❤️ They do have a time drift, though.

text/0015-time-monotonic.md
```diff
@@ -0,0 +1,188 @@
...
+* **Swift**: [`DispatchTime`] vs. [`DispatchTimeInterval`]
+* **Zig**: [`std.time.Instant`]
+
+# Unresolved questions
+
+- The name for the type is not settled. Alternatives:
+  - `Time::Monotonic`: Concise and explicit about clock domain, but less precise
+    about _what_ it is. Aligns with existing `Time.monotonic`.
+  - `Time::Instant`: More generic, leaves room for multiple backends.
+  - ?
```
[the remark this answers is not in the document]
straight-shoota: FWIW it can still be confusing because for example in Java, `Instant` represents an instant on the calendar time line.

text/0015-time-monotonic.md
```diff
@@ -0,0 +1,220 @@
...
+retaining the same logic.
+Existing uses should continue to work as before, despite the type change.
+
+```cr
+struct Time::Instant
+  include Comparable(self)
+
+  # Returns the current reading of the monotonic clock.
+  def self.now : self
+  end
```
straight-shoota: **suggestion:** Following the rename to `Time::Instant` (https://github.com/crystal-lang/rfcs/pull/15#discussion_r2346451070) I would suggest replacing this constructor with `Time.instant`. This is closer to the existing usage of `Time.monotonic`. It highlights the fact that we're not creating instances out of themselves, but they're rather readings from a clock (with `Time` providing access to that clock).

ysbaddaden: Definitely.

text/0015-time-monotonic.md
```diff
@@ -0,0 +1,219 @@
...
+Introduce a new type to represent instants on the monotonic timeline.
+This replaces the current [`Time.monotonic`] which returns [`Time::Span`].
+
+# Motivation
+
+Currently, `Time.monotonic` returns a `Time::Span`, which represents a
+*duration* rather than an *instant*. This overloading creates semantic
+ambiguity: a `Time::Span` type is overloaded. It can either mean "an elapsed
+duration" or "an absolute monotonic reading". This duplication can lead to
+confusion and accidental misuse.
```
ysbaddaden: Maybe:

```suggestion
`Time.monotonic` returns a `Time::Span` to represent an *instant* but a
`Time::Span` represents a *duration*, not an *instant*. This overloading
creates semantic ambiguity: a `Time::Span` type is overloaded. It can
either mean "an elapsed duration" or "an absolute monotonic reading".
This duplication can lead to confusion and accidental misuse.
```

text/0015-time-monotonic.md
```diff
@@ -157,17 +157,17 @@ while the system is suspended, some changes to the underlying clock are necessar
...
 | Platform | System call | Properties |
 |-|-|-|
-| Darwin | [`clock_gettime(CLOCK_UPTIME_RAW)`] | No adjustment, includes sleep, includes suspend? |
-| FreeBSD| [`clock_gettime(CLOCK_MONOTONIC_FAST)`][clock_gettime (FreeBSD)] | No adjustment, includes sleep, includes suspend, accurate to one timer tick |
-| Dragonfly | [`clock_gettime(CLOCK_MONOTONIC_FAST)`][clock_gettime (Dragonfly)] | No adjustment, includes sleep, includes suspend, accurate to one timer tick |
+| Darwin | [`clock_gettime(CLOCK_MONOTONIC)`][clock_gettime (darwin)] | No adjustment, includes sleep, includes suspend? |
+| FreeBSD| [`clock_gettime(CLOCK_MONOTONIC)`][clock_gettime (FreeBSD)] | No adjustment, includes sleep, includes suspend |
+| Dragonfly | [`clock_gettime(CLOCK_MONOTONIC)`][clock_gettime (Dragonfly)] | No adjustment, includes sleep, includes suspend |
 | Linux | [`clock_gettime(CLOCK_BOOTTIME)`][clock_gettime (Linux)] | Slewed, includes sleep, includes suspend |
 | NetBSD | [`clock_gettime(CLOCK_MONOTONIC)`][clock_gettime (NetBSD)] | ? |
```
ysbaddaden: From the man page:

```suggestion
| NetBSD | [`clock_gettime(CLOCK_MONOTONIC)`][clock_gettime (NetBSD)] | No adjustment, ? |
```

text/0015-time-monotonic.md
```diff
@@ -0,0 +1,261 @@
...
+| Darwin | `mach_absolute_time` (equivalent to [`clock_gettime(CLOCK_UPTIME_RAW)`][clock_gettime (darwin)]) |
+| Other Unix | `clock_gettime(CLOCK_MONOTONIC)` |
+| Windows | `QueryPerformanceCounter` |
+
+In order to implement the expectation of being as steady as possible and tick
+while the system is suspended, some changes to the underlying clock are necessary.
+
+| Platform | System call | Properties |
+|-|-|-|
+| Darwin | [`clock_gettime(CLOCK_MONOTONIC)`][clock_gettime (darwin)] | No adjustment, includes sleep, includes suspend? |
```
straight-shoota: **issue:** It looks like `CLOCK_MONOTONIC` on darwin has a resolution of 1 microsecond. I only noticed that experimentally because specs that need a higher resolution started failing (#16511).

So apparently we need to use `CLOCK_MONOTONIC_RAW` for a higher resolution (~42 nanoseconds).
```suggestion
| Darwin | [`clock_gettime(CLOCK_MONOTONIC_RAW)`][clock_gettime (darwin)] | No adjustment, includes sleep, includes suspend? |
```

PR #18, opened 2025-12-23
RFC 0018: Macro Methods

Blacksmoke16:
Preview: https://github.com/crystal-lang/rfcs/blob/1195640eaf4f6780bb02e1839a35552adb42d04e/text/0018-macro-methods.md
PoC PR: TODO

A proposal based on the OP from https://github.com/crystal-lang/crystal/issues/8835. Simple re-usable macro methods without the extra AST node monkey patching stuff.

### DIFF
--- a/text/0018-macro-methods.md
+++ b/text/0018-macro-methods.md
@@ -1,0 +1,227 @@
+---
+Feature Name: macro-methods
+Start Date: 2025-12-22
+RFC PR: "https://github.com/crystal-lang/rfcs/pull/18"
+Issue: "https://github.com/crystal-lang/crystal/issues/8835"
+---
+
+# Summary
+
+User-defined macro methods (`macro def`) allow reusable logic within macros, reducing code duplication in macro-heavy codebases. Unlike regular macros which generate code, macro methods operate purely within the macro expansion context and return AST nodes.
+
+# Motivation
+
+Currently, Crystal macros can become repetitive when similar logic is needed across multiple macro definitions.
+There's no way to extract common compile-time computations into reusable functions.
+This leads to:
+
+- Code duplication across macro definitions
+- Difficulty maintaining complex macro logic
+- No ability to compose compile-time operations
+
+Macro methods solve this by allowing developers to define reusable functions that operate on AST nodes during macro expansion, similar to their runtime counterparts.
+
+# Guide-level explanation
+
+## Defining Macro Methods
+
+A macro method is defined using `macro def` instead of `macro`:
+
+```crystal
+macro def format_name(name : StringLiteral) : StringLiteral
+  name.underscore.upcase
+end
+```
+
+Key differences from regular macros:
+
+1. The body is pure macro expressions, evaluated directly; no need for macro control characters within it `{{ }}` or `{% %}`
+2. Parameters and return types can specify expected AST node types
+3. The last expression is returned as an AST node
+
+## Calling Macro Methods
+
+Macro methods are called from within macro expansion contexts (`{{ }}` or `{% %}`):
+
+```crystal
+macro generate_constant(name)
+  CONST_{{ format_name(name) }} = "value"
+end
+
+generate_constant("myName")  # Generates: CONST_MY_NAME = "value"
+```
+
+## Type-Scoped Macro Methods
+
+Macro methods can be defined within types and called with that type as receiver:
+
+```crystal
+class Foo
+  macro def helper(items : ArrayLiteral) : ArrayLiteral
+    items.map { |x| x.stringify }
+  end
+end
+
+macro generate
+  {{ Foo.helper([1, 2, 3]) }}  # => ["1", "2", "3"]
+end
+```
+
+## Type Restrictions
+
+Type restrictions use AST node type names:
+
+```crystal
+macro def process(value : StringLiteral | SymbolLiteral) : StringLiteral
+  value.id.stringify
+end
+```
+
+Any AST node type can be used such as `StringLiteral`, `NumberLiteral`, `ArrayLiteral`, etc.
+`ASTNode` or no restriction may be used to to accept any type.
+
+## Blocks
+
+Macro methods support blocks via `&block`, or just `&` if it only yields and does not interact with the block parameter:
+
+```crystal
+macro def with_wrapper(&)
+  "before " + yield + " after"
+end
+
+{{ with_wrapper { "content" } }}  # => "before content after"
+```
+
+## Visibility
+
+Private macro methods cannot be called with an explicit receiver:
+
+```crystal
+class Foo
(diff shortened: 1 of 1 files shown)


ysbaddaden:
Some personal observations:

1. `ProcLitral#call` is the main alternative (and/or complement) and shall be documented here. See https://github.com/crystal-lang/crystal/issues/16451.

2. Because of the proximity with `def` I think we should define macro defs on explicit types, for example `macro def Foo.name` and `macro def self.name`.

Blacksmoke16:
Is the "return value" just whatever is the last line wrapped in `{{ }}`? What would the expected behavior be for a macro that is doing something like:
```cr
macro generate_constant(name)
  CONST_{{ name.id }}
end
```

If we want macros to support `next` and `return` keywords and such too would that conflict with using `{{ }}` as well?

BlobCodes:
> Is the "return value" just whatever is the last line wrapped in {{ }}? What would the expected behavior be for a macro that is doing something like (...)

  There is no explicit return value, the entire macro is executed as it works right now, its entire output is parsed into an ASTNode and this ASTNode is then returned to the caller.

  If the given example macro was called with the name "foo", it would return Path("CONST_foo").
  If the example macro was called with the name "A = 2", it would return `Assign(@target = Path("CONST_A"), @value = NumberLiteral(2))`

> If we want macros to support next and return keywords and such too would that conflict with using {{ }} as well?

  Yes, explicit returns would probably conflict with the current template syntax.

  That's something that could be explored with macro defs.

> Is the expectation that if you call one of these macros in a place its not expected to be, that it just fails or works but with undefined behavior?

  Why should a macro result in undefined behaviour if it's called within a nested context?

Blacksmoke16:
> Why should a macro result in undefined behaviour if it's called within a nested context?

I was mainly thinking about my use case where some of my `macro defs` would be like:

```cr
macro foo(x)
  {%
    # Interacting with constants and stuff
    # Intended to be called within a macro expression at certain places in the code
  %}
end
```
So if you tried to use this macro outside of a macro expression where you're expected for it to run, what would it do? Probably error that the const is unknown? Maybe not a big deal in practice, but this is kinda why i liked `macro def` being a separate thing. Just easier, for me at least, mentally to know that "okay these macros i use in a macro context while this other kind of macro is for this other context" versus needing to possibly handle the case where a macro is used in a place it wasn't expected to be.

Blacksmoke16:
> otherwise I think you'd end up with a lot of template macros that just call a macro def and splat the result into the program, ie:

What would the use case/reasoning for this be? Because in this example you don't even need the `macro def` for it to work:
```cr
macro foo(args)
  {{ "from macro method: #{args.stringify}" }}
end

foo "hello!"
```

willhbr:
I should've been clearer, I mean if I have a `macro def` that either does something you can't do with a normal macro, or is just a macro def because that's how I wanted to write it, to expose an API where it can be called like `foo("some value")` instead of `{{ foo("some value") }}` I'd have to wrap it in a regular macro and forward the arguments.

willhbr:
If a normal/template macro doesn't have a return type, would you still be able to call it from a macro def?

I don't know if this would actually end up being useful—it's hard to know without using it—but I think it would be useful to allow extracting parts of templates into separate methods to make code more readable/testable/etc.

I thought through rewriting my [dependency injector macro](https://codeberg.org/willhbr/geode/src/branch/main/src/geode/dependency_injector.cr) if I had macro methods, and I came to this section:

```crystal
def self.inject(scope)
  {{ @type }}.new(
    {% for arg in method.args %}
      {% if arg.restriction.nil?
           arg.raise "needs restriction on #{arg}"
         end %}
      {% if arg.restriction.resolve?.nil?
           arg.raise "Unable to resolve #{arg.restriction}"
         end %}
      {% if arg.restriction.resolve < Geode::Injector::Provider %}
        {{ arg.restriction }}.inject(scope),
      {% elsif arg.restriction.resolve < Geode::Injector::Lazy %}
        {{ arg.restriction }}.inject(scope),
      {% else %}
        scope.get({{ arg.restriction }}),
      {% end %}
    {% end %}
  )
end
```

I could pull out the verification into a separate macro def easily:

```crystal
macro def validate_injectable_arg(arg)
  if arg.restriction.nil?
    arg.raise "needs restriction on #{arg}"
  end
  if arg.restriction.resolve?.nil?
    arg.raise "Unable to resolve #{arg.restriction}"
  end
end
```

But I'm not sure if I could do the same for the code that generates the actual code in the loop. It would want to be a template macro, so I could do this:

```crystal
macro build_creation(arg, scope)
  %scope = {{ scope }}
  {% if arg.restriction.resolve < Geode::Injector::Provider %}
    {{ arg.restriction }}.inject(%scope),
  {% elsif arg.restriction.resolve < Geode::Injector::Lazy %}
    {{ arg.restriction }}.inject(%scope),
  {% else %}
    %scope.get({{ arg.restriction }}),
  {% end %}
end
```

Then the code generated by the main macro would look like this:

```crystal
SomeInjectableType.new(
  build_creation(some_arg : HTTP::Context),
  build_creation(another_arg : User),
)
```

Which I guess is fine, but it does mean we've got another layer of macro expansion that needs to happen, and it's a bit harder to see what the final generated code is (`{% debug %}` won't show the fully expanded code in this case, right?).

It would be nice to be able to call any macro from any other macro and have the result inserted into the code, so I could do:

```crystal
def self.inject(scope)
  {{ @type }}.new(
    {% for arg in method.args %}
      {% validate_injectable_arg arg %}
      {{ build_creation(arg, MacroId.new(:scope)) }},
    {% end %}
  )
end
```

Would I be able to create and return an actual AST from `build_creation`, without building a `Crystal::Macros::Call` myself, and then call that from another macro?

zw963:
perfer `macro def` way, I can search this easier use rg for which macro is defined for another macros, it for readability.

Introduce this is for DRY when macro, eliminate duplicate code, but, I don't see any need for a block at this point—this is certainly not something that should be on the agenda for the current version.

(Translated with qwen3.5:35b)

Blacksmoke16:
> Would I be able to create and return an actual AST from `build_creation`, without building a `Crystal::Macros::Call` myself, and then call that from another macro?

I been thinking about this and that's a fair point. I'd have to see how this works in my POC branch. My current thinking is it would be the same if there is just a single ASTNode being returned, but having a macro able to inline the output of another macro feels like a different but related use case. You can of course expand a macro as part of the output of another macro so could maybe do something like:
```cr
def self.inject(scope)
  {{ @type }}.new(
    {% for arg in method.args %}
      {% validate_injectable_arg arg %}
      build_creation({{ arg }}, :scope),
    {% end %}
  )
end
```
Where `build_creation` is a normal `macro` that would expand as if it was inlined.

PR #19, opened 2026-02-02
RFC: Immutable ENV

ysbaddaden:
Preview:

https://github.com/crystal-lang/rfcs/blob/rfc-immutable-env/text/0019-immutable-env.md

Related to:

- https://forum.crystal-lang.org/t/eliminate-environment-modifications/8533
- https://github.com/crystal-lang/crystal/issues/16449
- https://github.com/crystal-lang/crystal/pull/16567
- https://github.com/crystal-lang/crystal/pull/16591
- https://github.com/crystal-lang/crystal/pull/16601

### DIFF
--- a/text/0019-immutable-env.md
+++ b/text/0019-immutable-env.md
@@ -67,10 +67,6 @@ identified:
   methods, though sometimes it might no be easy (e.g. tweak `PATH` to test
   sub-processes).
 
-- Testing `ENV` itself is challenging. Maybe we could compile an additional
-  program (once), then each test would call it with a custom environment and a
-  list of operations to execute and assert the output.
-
 After running a survey, the main case for mutating `ENV` is to load a `.env`
 file (or equivalent), have it mutate `ENV` and then have anything that takes
 environment variables just use it. This behavior will no longer be possible.


ysbaddaden:
I'm personally leaning on the startup snapshot (Java / Swift / Go) + explicitly unsafe method to set a variable (Rust).

We're safe by default. We encourage to never set a variable, but developers can choose to be unsafe at their own perils (you've been warned).

text/0019-immutable-env.md
```diff
@@ -0,0 +1,199 @@
...
+  list of operations to execute and assert the output.
+
+After running a survey, the main case for mutating `ENV` is to load a `.env`
+file (or equivalent), have it mutate `ENV` and then have anything that takes
+environment variables just use it. This behavior will no longer be possible.
+
+Applications could make sure to source any `.env` before the program is executed
+through a wrapper shell script, shell plugin ([direnv]), the `--env-file`
+argument to docker and podman, ... or develop alternatives to not depend on
+`ENV` only.
```
straight-shoota: **suggestion:** I would move up alternatives to `ENV` as primary recommendation. `ENV` is great for ingesting external configuration. But when you load configuration data directly in the process, there are much better alternatives.

For example, class variables are type safe and can be easily documented.
If you really need it, you can store free-form key-value mappings in a custom `Hash(String, String)`, accessible as a constant.
More complex configuration libraries can help as well.

Prepping the environment before executing the application can work well, but it's a deployment problem, not an engineering solution.

text/0019-immutable-env.md
```diff
@@ -0,0 +1,199 @@
...
+
+> [!NOTE]
+> We can't recommend enough a configuration library such as [totem] to revisit
+> and centralize your application's settings.
```
straight-shoota: **thought:** I'm not sure about such a strong endorsement for a specific library.
I personally have no experience with totem. Maybe it is _that_ good?
It doesn't seem to be used a lot, though.

text/0019-immutable-env.md
```diff
@@ -0,0 +1,199 @@
...
+  argument of `Process.new`.
+
+- Testing the program's behavior around `ENV` variables. We could mock `ENV`
+  methods, though sometimes it might no be easy (e.g. tweak `PATH` to test
+  sub-processes).
+
+- Testing `ENV` itself is challenging. Maybe we could compile an additional
+  program (once), then each test would call it with a custom environment and a
+  list of operations to execute and assert the output.
+
```
ysbaddaden:
```suggestion
```

straight-shoota:
Before settling on anything final, we should do a bit more exploration to better understand the cases where there might be a need for environment modification.

Deprecating the existing methods would help to raise awareness and identify affected code. But then it's not great to issue deprecation warnings without clarity on how a replacement might look like.

We should just start with grepping through source code to identify `ENV` mutations in the ecosystems and help library authors to replace them. For engaging in discussions it would be nice to collect a document with a guide-level explanation and descriptions for common patterns. I'm not sure if an RFC is the best place for that, but it might do fine.
For this purpose we would need to merge the RFC in order to create a stable URL to link to. Even if it's not final. So maybe we should consider this a preliminary RFC which describes the problem and offers workarounds. As such it can grow over time as we discover more cases to document, for example. At some point we start a new RFC for deciding on whehter and how mutability stays in the stdlib.

I've already began some identifications, and I've deployed on a yak shaving tour with [`shards`](https://github.com/crystal-lang/shards).

ysbaddaden:
I checked my own projects: I already only read from `ENV` in actual application code. 

For example instead of `ENV["APP_ENV"] = "test"` I use an explicit setup like `App.setup(env: "test")` (reminiscent of `Log.setup_from_env`), and the same project uses an external config files with overrides from `ENV` (read-only).

The few cases where I write to `ENV` are all in test suites.

For example tests for methods that use `ENV` (like the configurator above) that could just use mocking, which makes me think an official `ENV.mock` might be nice:

```crystal
ENV.mock({ "KEY" => "value" }) do
  ENV["KEY"]?.should eq("value")
end
```

Other cases are to cleanup `ENV` for the duration of the test suite (a docker client) that could just use mocking too, or use globals like:

```crystal
# foo.cr
module Foo
  class_properties(value) { ENV["DEFAULT"]? }
end

# spec/spec_helper.cr
require "foo"
Foo.value = "explicit"

# spec:
Foo.value.should eq("explicit")
```

Edit: _and these examples should be in the guide level section_.

PR #20, opened 2026-02-17
RFC 0020: `@[TargetFeature]` annotation

straight-shoota:
**Preview:** https://github.com/crystal-lang/rfcs/blob/rfc/target-annotation/text/0020-target-annotation.md

### DIFF
--- a/text/0020-target-annotation.md
+++ b/text/0020-target-annotation.md
@@ -34,16 +34,18 @@ A runtime check selects the appropriate implementation.
 
 ```cr
 @[Target(features: "+sve,+sve2")]
-def foo_sve2
+private def foo_sve2
 end
 
-def foo_portable
+private def foo_portable
 end
 
-if cpu_supports_sve?
-  foo_sve2
-else
-  foo_portable
+def foo
+  if cpu_supports_sve?
+    foo_sve2
+  else
+    foo_portable
+  end
 end
 ```
 
--- a/text/0020-target-annotation.md
+++ b/text/0020-target-annotation.md
@@ -135,7 +135,6 @@ Several modern systems languages provide mechanisms to compile code using CPU fe
 ## Unresolved questions
 
 - Should there be warnings and should they be be suppressible or configurable?
-- Do we really need to guard the `@[Target]` def with arch flags or only the callsites?
 
 
 ## Future possibilities
--- a/text/0020-target-annotation.md
+++ b/text/0020-target-annotation.md
@@ -22,7 +22,7 @@ The current compiler refuses to generate instructions that are not supported by 
 That means every compilation can target exactly one configuration which can either be generic and miss out on advanced features or depend on advanced features and reduce portability.
 This prohibits legitimate use cases for having optional code paths, depending on feature support at runtime.
 
-You thus need to tell LLVM to enable CPU features scoped to specific functions only, which sounds reasonable. Annotations also sound good.
+You thus need to tell LLVM to enable CPU features scoped to specific functions only. In this proposal we suggest an annotation for it.
 
 
 ## Guide-level explanation
--- a/text/0020-target-annotation.md
+++ b/text/0020-target-annotation.md
@@ -47,6 +47,9 @@ else
 end
 ```
 
+> [!CAUTION]
+> | This is an unsafe feature and programs may crash with an invalid instruction error at runtime if CPU feature checks are inaccurate.
+
 
 ## Reference-level explanation
 
--- a/text/0020-target-annotation.md
+++ b/text/0020-target-annotation.md
@@ -140,8 +140,12 @@ Several modern systems languages provide mechanisms to compile code using CPU fe
 
 ## Future possibilities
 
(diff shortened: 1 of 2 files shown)


text/0020-target-annotation.md
```diff
@@ -0,0 +1,136 @@
...
+- Clang / GCC have similar function annotations `__attribute__((target("avx2")))`/`[[ gnu::target("avx2") ]]`.
+- Go doesn't validate instructions in its own assembly language.
+- Zig requires separate code + object files each compiled separately with the correct flags and then the object files are linked. There is an ongoing discussion in [zig #1018] with a similar motivation as this RFC.
+
+
+## Unresolved questions
+
+- Should there be warnings and should they be be suppressible or configurable?
+- Do we really need to guard the `@[Target]` def with arch flags or only the callsites?
+- How can we provide helpers for  runtime feature detection?
```
stakach: Rust provides a helper library which is somewhat similar to this
https://github.com/spider-gazelle/simd/blob/main/src/simd/detect.cr

text/0020-target-annotation.md
```diff
@@ -0,0 +1,136 @@
...
+Feature and CPU selections only accept values valid for the current target platform architecture.
+That means their definition may typically need guard clauses.
+
+```cr
+{% if flag?(:aarch64) %}
+  @[Target(features: "+avx2")]
+  def foo_avx2
+  end
+{% end %}
+```
```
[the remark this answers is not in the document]
beta-ziliani: Fair point!

stakach (approved):

text/0020-target-annotation.md
```diff
@@ -0,0 +1,136 @@
...
+
+It supports these parameters:
+
+- `feature`: Select specific platform architecture features. See https://llvm.org/doxygen/classllvm_1_1SubtargetFeatures.html
+  ```cr
+  @[Target(features: "+avx2")]
+  def foo_avx2
+  end
+  ```
+- `cpu`: Select a specific CPU type.
```
ysbaddaden: Maybe:
```suggestion
- `cpu`: Select a specific CPU model.
```

text/0020-target-annotation.md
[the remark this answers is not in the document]
ysbaddaden: This is acceptable as per Crystal semantics: the function won't be considered until call. Nothing wrong with wrapping the definition, so the function won't be parsed into an AST.

text/0020-target-annotation.md
```diff
@@ -0,0 +1,136 @@
...
+
+When a method has been compiled with a feature and that feature is not supported on the platform at runtime, calling this method is undefined behaviour.
+
+Method with `@[Target]` feature annotation may not be inlined into an incompatible context.
+
+
+## Drawbacks
+
+- Increased complexity in code generation and target feature management.
+- Risk that binaries may crash at runtime when this feature is handled unsafely.
```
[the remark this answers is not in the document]
ysbaddaden: I would phrase "This is an unsafe feature and programs may crash with an invalid instructions error" when CPU checks aren't thorough.

text/0020-target-annotation.md
```diff
@@ -0,0 +1,142 @@
...
+Feature Name: target_annotation
+Start Date: 2026-02-17
+RFC PR: "https://github.com/crystal-lang/rfcs/pull/0020"
+Issue: "https://github.com/crystal-lang/crystal/issues/16570"
+Implementation PR: "https://github.com/crystal-lang/crystal/pull/16571"
+---
+
+## Summary
+
+Introduces a new `@[Target]` annotation that allows setting codegen options for individual methods.
```
straight-shoota: **thought:** I'm wondering if the name `Target` is too generic. Maybe we should use more specific name for the annotation? For example `TargetFeature` (akin to Rust's `target_feature`).

stakach: `@[Target(features: "+sve,+sve2", cpu: "apple-m1")]`
vs
`@[TargetFeature(features: "+sve,+sve2", cpu: "apple-m1")]`
or maybe
```
@[TargetFeature("+sve,+sve2")]
@[TargetCPU("apple-m1")]
@[NoDebugSymbols]
@[Optimize(None)] / @[Optimize(Size)]
```

stakach: clang also [supports this](https://releases.llvm.org/7.1.0/tools/clang/docs/AttributeReference.html#:~:text=cpu_dispatch%20(clang::cpu_dispatch%2C%20clang::cpu_dispatch),-Supported%20Syntaxes&text=The%20result%20of%20this%20is,the%20new%20processor%20can%20execute.&text=A%20dispatching%20(or%20resolving)%20function,are%20permissible%20for%20ICC%20compatibility.&text=Note%20that%20it%20is%20possible,a%20multiversioned%20function%20is%20undefined.)

ysbaddaden: The micro-architecture argument is much better than a specific cpu model. For the later, you should use the `--mcpu` flag to build an executable for the specific target (a RPI), while a micro-architecture specifies a minimal set of features.

That being said, it will make runtime detection harder (and slower), especially automatic dispatch, and maybe even wrong as we might require more features than the function needs.

ysbaddaden (approved):
This PR is ready for implementation. Poke @stakach.

beta-ziliani (approved):

PR #25, opened 2026-03-03
RFC 0025: Modern Process Spawn API

straight-shoota:
**Preview**: https://github.com/crystal-lang/rfcs/blob/rfc/process-spawn-api/text/0025-process-spawn-api.md


text/0025-process-spawn-api.md
```diff
@@ -0,0 +1,306 @@
...
+The `Process.capture` group of methods provides a convenient tool to capture the output of the subprocess.
+
+By default, `Process.capture` captures both `stdout` and `stderr`, but it only returns `stdout`.
+The captured output of `stderr` is passed to the error message in case the process was unsuccessful. If the error output is unreasonably long, it's truncated to keep only the first and last 32kB in order to prevent resource exhaustion.
+Passing any value other than `error: :pipe` prevents capturing the error stream.
+
+- `Process.capture(...) : String`: Returns captured stdout and raises if the process terminates unsuccessfully.
+- `Process.capture?(...) : String?`: Returns captured stdout or `nil` if the process terminates unsuccessfully.
+- `Process.capture_result(...) : Process::Result`: Returns captured result and raises if the process terminates unsuccessfully.
+- `Process.capture_result?(...) : Process::Result?`: Returns captured result or `nil` if the process terminates unsuccessfully.
```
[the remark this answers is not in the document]
ysbaddaden: I think what's important is for run vs capture to have the same behavior, to avoid inconsistencies. If we can get better names, too, then that's a nice to have.

Note: maybe `.try_run? : Status?` would be more explicit than `.run?` (same for `.try_capture? : Result?`), but that still makes `.capture? : String?` a bit inconsistent — even if there's no `.run?`

text/0025-process-spawn-api.md
```diff
@@ -0,0 +1,306 @@
...
+The `Process.capture` group of methods provides a convenient tool to capture the output of the subprocess.
+
+By default, `Process.capture` captures both `output` and `error`, but it only returns `output`.
+The captured output of `error` is passed to the error message in case the process was unsuccessful. If the error output is unreasonably long, it's truncated to keep only the first and last 32kB in order to prevent resource exhaustion.
+Passing any value other than `error: :pipe` prevents capturing the error stream.
+
+- `Process.capture(...) : String`: Returns captured output and raises if the process terminates unsuccessfully.
+- `Process.capture?(...) : String?`: Returns captured output or `nil` if the process terminates unsuccessfully.
+- `Process.capture_result(...) : Process::Result`: Returns captured result and raises if the process terminates unsuccessfully.
+- `Process.capture_result?(...) : Process::Result?`: Returns captured result or `nil` if the process terminates unsuccessfully.
```
straight-shoota: **issue:** I just realized I did a mistake here and this is not how it's supposed to be.

`capture_result` should not raise on an unsuccessful exit status. This method is specifically for the use case where you want to handle different exit status values.

```suggestion
- `Process.capture(...) : String`: Returns captured output or raises if the process does not terminate successfully.
- `Process.capture?(...) : String?`: Returns captured output or `nil` if the process does not terminate successfully.
- `Process.capture_result(...) : Process::Result`: Returns captured result or raises if the process fails to execute.
- `Process.capture_result?(...) : Process::Result?`: Returns captured result or `nil` if the process fails to execute.
```

We could consider `Process.capture_result!` as a future enhancement that raises if the process terminates unsuccessfully. It would have the benefit over `.capture` that it captures both `stdout` and `stderr`. But not sure if that's relevant enough.

text/0025-process-spawn-api.md
```diff
@@ -0,0 +1,306 @@
...
+Many languages' standard libraries represent a command and its arguments as a single list.
+
+Go provides [`Cmd.Output()`](https://pkg.go.dev/os/exec#Cmd.Output) to capture process output. It always captures stderr and exposes it in the error value. The captured amount is limited to preserving only the first 32k and last 32k bytes (see [`os.exec.ExitError`](https://pkg.go.dev/os/exec#ExitError)).
+
+## Unresolved questions
+
+- Deprecation of legacy API methods.
+- Deprecation of command literals.
+
+## Future possibilities
```
straight-shoota, line 307: **thought:** I'm wondering about the details of handling newlines. Maybe this 
Naturally, I'm not very inclined to the idea of altering the captured string. But I'm willing to give it a minute to think about in the interest of convenience.

Line breaks are different on different operating systems (simplified: Unix LF vs Windows CRLF).
In the prototype application of `Process.capture` for `shards` (https://github.com/crystal-lang/shards/pull/701), this is a serious concern in particular for portability of specs.
I ended up normalizing line endings explicitly in a project-local helper method:
 
https://github.com/straight-shoota/shards/blob/c5b21b143fb4a6afa0ef28db13a15fbc4716f585/spec/support/factories.cr#L360-L368

This is not an ideal solution though, considering that many other libraries face similar issues for expectations about the output from other processes.
I'm sure it would not be wise to implicitly normalize line breaks. But maybe there could be an explicit normalizer? `Process::Result#normalize` perhaps?
And/or since this is especially important for specs, maybe we could provide a spec helper for this?

To be fair: Shards specs currently have 3 examples with line breaks in the expected text. Two of them only test a single line, so there is currently only one multi-line expectation.

https://github.com/crystal-lang/shards/blob/d1f61d36497982fcbd2d68e9f9464c52788f4883/spec/integration/install_spec.cr#L898-L903

This is not huge. But I figure it might be useful to have more of those expectations. And for that we need an easy solution to make it convenient to use.

straight-shoota, line 307: **thought:** Another thought about newlines: When capturing output that consists of a single word (or similar) you typically want to remove a trailing newline.
So captured process output is quite often transformed with `.chomp`.

Maybe it's fine to have this explicit call.
But I'm also thinking maybe it could be more convenient to not require another call?

Would it be feasible for `.capture` to chomp implicitly? Maybe only when the output is a single line (i.e. `index('\n') == size -1`)?

As mentioned above, I'm very sceptical of altering the captured string. But maybe there is something in it?

ysbaddaden: The single multiline case in Shards can be rewritten to test all 3 lines independently, then we won't need to normalize anymore.

Now, that doesn't invalidate such a method, but `#normalize` might be too generic—what is it normalizing?

ysbaddaden: That should never be implicit: we know nothing about the output. It might not even be plain text but binary.

Sija: I'd suggest adding a `chomp` parameter.

straight-shoota: I don't think a parameter would have much benefit over explicitly calling `chomp` on the result.
Perhaps we could potentially avoid the intermediary string allocation. But that doesn't work out of the box and would require support from `IO#gets_to_end` (or a custom implementation).

straight-shoota:
I added splat parameter overloads for all new `Process` API methods, as suggested in https://github.com/crystal-lang/crystal/pull/16780#issuecomment-4117513058

text/0025-process-spawn-api.md
Sija: True, although I thought it would make sense as there's already `gets(chomp = true)`.

text/0025-process-spawn-api.md
```diff
@@ -0,0 +1,319 @@
...
+### Single `args` array
+
+The entire command line consisting of the command and its arguments is represented as a single value which can be passed around easily.
+
+String array literals are a convenient way to write a command line. The syntax reads similar to a shell command, but it's actually an array and thus avoids shell parsing rules.
+
+```cr
+# splat parameter
+Process.run("crystal", "tool", "format", path)
+Process.run("crystal", "tool", "format", *paths) # `paths` must be a Tuple
```
straight-shoota, line 96: **aside:** Looking at this, I'm wondering if we could make splat parameters work with any collection type. It would mean the arguments cannot be represented as a tuple. It would require a dynamically sized container type. 🤔
The use case is not super strong of course, but it might be a bit of a nuisance.

zw963:
remove shell: true is okay,  but, I think we really should not remove the ```::system``` and  \`...` in the future, that's too radical, many user like use it just for convenience,  We shouldn't just remove the ways most users are accustomed to, simply to improve Windows support or make the interface look more uniform。

So what if it only supports Linux? Not every command needs to be perfectly supported on both Windows and Linux.

ysbaddaden:
We plan to obsolete `system` and the command operator because they're unsafe, not portable and bad overall.

`Process.run %W[]` and `Process.capture %W[]` are simpler (no extra shell), safer (no arg injection), portable (no `sh` vs `cmd`) and better overall, without being much worse to type and read.

straight-shoota:
The future of `::system()` and ``::`()`` are explicitly not part of this RFC. Let's discuss that in a follow-up.

text/0025-process-spawn-api.md
straight-shoota: https://forum.crystal-lang.org/t/idea-splat-def-parameters-into-array/8898

